PR #19, opened 2018-07-11
Timeout and Lambda Proxy Integration support

kobalski:
- Timeout detection support
- Parse Lambda Proxy Integration response for error detection for 5xx and 4xx responses

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Check out the [configuration part](https://thundra.readme.io/docs/nodejs-configu
 | thundra_lambda_warmup_warmupAware        |  bool  |     false     |
 | thundra_lambda_trace_request_skip        |  bool  |     false     |
 | thundra_lambda_trace_response_skip       |  bool  |     false     |
+| thundra_lambda_timeout_margin            | number |     200       |
 | thundra_lambda_publish_rest_baseUrl      | string |  https<nolink>://collector.thundra.io/api  |
 
 #### 2. Module initialization parameters
--- a/src/constants.js
+++ b/src/constants.js
@@ -15,5 +15,13 @@ export const URL = url.parse(
         : 'https://collector.thundra.io/api'
 );
 
+export class TimeoutError extends Error {
+    constructor(args) {
+        super(...args);
+        Error.captureStackTrace(this, TimeoutError);
+        Object.setPrototypeOf(this, TimeoutError.prototype);
+    }
+}
+
 export const PROC_STAT_PATH = '/proc/self/stat';
 export const PROC_IO_PATH = '/proc/self/io';
--- a/src/thundra-wrapper.js
+++ b/src/thundra-wrapper.js
@@ -18,6 +18,7 @@
 
 import uuidv4 from 'uuid/v4';
 import Reporter from './reporter';
+import {TimeoutError} from './constants';
 
 class ThundraWrapper {
     constructor(self, event, context, callback, func, plugins, pluginContext, apiKey) {
@@ -49,6 +50,8 @@ class ThundraWrapper {
                 });
             }
         };
+
+           this.timeout = this.setupTimeoutHandler(this);
     }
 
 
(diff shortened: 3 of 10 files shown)


src/thundra-wrapper.js
```diff
@@ -100,19 +103,75 @@ class ThundraWrapper {
...
                 response: result
             };
+
+            const skipParseResponse = process.
+                env.thundra_lambda_skip_parse_response === 'true' ? true : false;
+
+            if (skipParseResponse || this.isErrorResponse(result)) {
+                afterInvocationData = {
+                    error: new Error("Lambda returned with error response."),
+                    response: null
```
serkan-ozal: Assign `result` to `response` event though there is error. So user can see which status code was returned

src/thundra-wrapper.js
```diff
@@ -100,19 +103,75 @@ class ThundraWrapper {
...
                 error: error,
                 response: result
             };
+
+            const skipParseResponse = process.
+                env.thundra_lambda_skip_parse_response === 'true' ? true : false;
+
+            if (skipParseResponse || this.isErrorResponse(result)) {
+                afterInvocationData = {
+                    error: new Error("Lambda returned with error response."),
```
serkan-ozal: We are checking HTTP error codes (4xx and 5xx) therefore we can use specific error type for this case such as `HTTPError`. So user can detect it in the UI. 

Also it would be good to have status code in the error message.

kobalski: 👍

src/thundra-wrapper.js
kobalski: 👍

src/thundra-wrapper.js
```diff
@@ -100,19 +103,75 @@ class ThundraWrapper {
...
             this.reported = true;
-            const afterInvocationData = {
+
+            let afterInvocationData = {
                 error: error,
                 response: result
             };
+
+            const skipParseResponse = process.
+                env.thundra_lambda_skip_parse_response === 'true' ? true : false;
```
serkan-ozal: Since this check is HTTP response specific, I think we can use `thundra_lambda_http_responseCheck_skip` as environment variable. 

The reason why I have used this naming convention is that 
- Later there might be other environment variables related with HTTP domain in `thundra_lambda_http_...` format
- Later there might be other environment variables related with HTTP response check (not just about skipping but customizing check behaviour) in `thundra_lambda_http_responseCheck...` format

src/constants.js
```diff
@@ -15,5 +15,26 @@ export const URL = url.parse(
         : 'https://collector.thundra.io/api'
 );
 
+export class TimeoutError extends Error {
+    constructor(message) {
+        super();
+        this.message = message;
+        this.name = 'TimeoutError';
+        Error.captureStackTrace(this, TimeoutError);
+        Object.setPrototypeOf(this, TimeoutError.prototype);
```
serkan-ozal: just for curiosity, why setting prototype is needed here?

README.md
```diff
@@ -57,6 +57,7 @@ Check out the [configuration part](https://thundra.readme.io/docs/nodejs-configu
 | thundra_lambda_warmup_warmupAware        |  bool  |     false     |
 | thundra_lambda_trace_request_skip        |  bool  |     false     |
 | thundra_lambda_trace_response_skip       |  bool  |     false     |
+| thundra_lambda_timeout_margin            | number |     200       |
```
serkan-ozal: HTTP response check skip related environment variable should also be added here with default value is `false`

src/constants.js
kobalski: To pass the tests. During the transpilation when running tests `extends` keyword did not work. So I added setPrototypeOf to ensure inheritance. Tests passed after `setPrototypeOf`.

src/thundra-wrapper.js
kobalski: 👍

README.md
kobalski: 👍

serkan-ozal: ok I see :+1:

PR #21, opened 2018-07-17
Move repo to typescript and implement manual instrumentation with opentracing

kobalski:


src/opentracing/Tracer.ts
```diff
@@ -0,0 +1,143 @@
...
+  wrapper<T extends (...args: any[]) => any>(spanName: string, func: T): T {
+    // tslint:disable-next-line:no-angle-bracket-type-assertion
+    return <T> ((...args: any[]) => {
+      try {
+        this.startSpan(spanName);
+        const returnValue = func(...args);
+        this.finishSpan();
+        return returnValue;
+      } catch (error) {
+        this.finishSpan();
```
serkan-ozal: shouldn't we throw error here?

src/opentracing/Tracer.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  destroy() {
+    this.recorder.destroy();
+    this.activeSpans = null;
+  }
+
+  wrapper<T extends (...args: any[]) => any>(spanName: string, func: T): T {
+    // tslint:disable-next-line:no-angle-bracket-type-assertion
+    return <T> ((...args: any[]) => {
```
serkan-ozal: I think, we should get active span and store it to a local variable here. Then while starting new span, the new span is attached to that stored span instead of active span while running callback. 

So with this way, callbacks can be attached to the spans which are active while creating callbacks.

src/opentracing/Tracer.ts
```diff
@@ -0,0 +1,143 @@
...
+    const parentContext = getParent(fields.references);
+
+    if (parentContext && !this.activeSpans.get(parentContext.spanId)) {
+        throw new Error('Invalid spanId : ' + parentContext.spanId);
+    }
+
+    if (fields) {
+      span = new ThundraSpan(this, {
+        operationName: fields.operationName || name,
+        parent: getParent(fields.references),
```
serkan-ozal: no need to get parent context again. it is already retrieved above and assigned to `parentContext`

src/opentracing/Tracer.ts
```diff
@@ -0,0 +1,143 @@
...
+import * as opentracing from 'opentracing';
+import ThundraSpan, { SpanEvent } from './Span';
+import ThundraSpanContext from './SpanContext';
+import ThundraRecorder from './Recorder';
+import ThundraSampler from './Sampler';
+import Utils from '../plugins/Utils';
+
+class ThundraTracer extends Tracer {
+  tags: any;
+  recorder: ThundraRecorder;
```
serkan-ozal: When there is custom recorder (other than `ThundraTracer`), does this cause exception while assigning?

src/opentracing/Tracer.ts
```diff
@@ -0,0 +1,143 @@
...
+  sampler: ThundraSampler;
```
serkan-ozal: When there is custom sampler (other than `ThundraSampler`), does this cause exception while assigning?

src/opentracing/Tracer.ts
```diff
@@ -0,0 +1,143 @@
...
+        tags: Object.assign(tags, this.tags, fields.tags),
+        startTime: fields.startTime,
+        rootTraceId,
+        className: fields.className,
+        domainName: fields.domainName,
+      });
+    } else {
+      span = new ThundraSpan(this, {
+        operationName: name,
+        parent: this.getActiveSpan(),
```
serkan-ozal: `parent` field above in the `if` statement is `ThundraSpanContext ` typed but as far as I see, in here, it is `ThundraSpan` typed (`ThundraSpan` typed `value` field of `SpanTreeNode`). 

So seems that at different execution paths, we pass different typed values for `parent` field to `ThundraSpan`

kobalski: Yes, you are right it should be 'this.getActiveSpan().spanContext'. It did not cause any problem, because we are tracking parent/child relation in the recorder. In our current structure, we do not use parent context.  This field will be beneficial when user sends data with custom recorder.

src/opentracing/Tracer.ts
kobalski: 👍 Fixed.

src/opentracing/Tracer.ts
kobalski: 👍

src/opentracing/Tracer.ts
kobalski: 👍

src/opentracing/Tracer.ts
kobalski: No, our code is transpiled to JS and JS has no types. In the config, user should provide an JS object with `record` method. I will give more detailed example in docs.

src/opentracing/Tracer.ts
kobalski: No, our code is transpiled to JS and JS has no types. In the config, user should provide an JS object with `isSampled` method. I will give more detailed example in docs.

PR #23, opened 2018-07-31
Implement Automatic Instrumentation

kobalski:


src/plugins/config/TraceConfig.ts
```diff
@@ -0,0 +1,55 @@
...
+    parseTraceDefEnvVariable(value: string): TraceOption {
+        const args = value.substring(value.indexOf('['), value.indexOf(']')).split(',');
+        const pattern = value.substring(0, value.indexOf('['));
+        const option: any = new TraceOption(pattern);
+        for (const arg of args) {
+            const tupple = arg.split('=');
+            const argKey = tupple[0];
+            const argValue = tupple[1];
+            if (option[argKey]) {
+                option[argKey] = (argValue === 'true');
```
serkan-ozal: Even though, current option values are `boolean`, there might be other value types (`operationName`, `serviceName`, ...). So I think it is better accept all option values as `string` and convert expected type at the caller side which uses option value

src/opentracing/instrument/ModuleHack.ts
```diff
@@ -0,0 +1,288 @@
...
+
+    shouldTrace(relPath: string): TraceOption {
+        try {
+            if (relPath.includes('node_modules')) {
+                return null;
+            }
+            for (const traceOption of this.traceConfig.traceDef) {
+                const patterns = traceOption.pattern.split(TRACE_DEF_SEPERATOR);
+                const regStr = patterns.slice(0, patterns.length - 1).join().replace(TRACE_DEF_SEPERATOR, '/');
+                const regExp = new RegExp(regStr);
```
serkan-ozal: We should cache `regExp` at somewhere (maybe in the `traceOption`). Pattern should not be parsed every time for trace check.

src/opentracing/instrument/ModuleHack.ts
```diff
@@ -0,0 +1,288 @@
+const Module = require('module');
+const falafel = require('falafel');
+const util = require('util');
+const path = require('path');
+
+const TRACE_ENTRY = 'var __njsEntryData__ = __njsTraceEntry__({file: %s, name: %s, line: %s, args: %s, argNames:%s});';
```
serkan-ozal: `file` and `line` parameters are not needed

src/opentracing/instrument/ModuleHack.ts
```diff
@@ -0,0 +1,288 @@
...
+        };
+
+        global.__njsTraceExit__ = function (args: any) {
+            try {
+                const span = self.tracer.getActiveSpan();
+                console.log(args.exceptionValue);
+                if (!args.exception) {
+                    span.setTag(RETURN_VALUE_TAG_NAME, new ReturnValue(typeof args.returnValue, args.returnValue));
+                } else {
+                    span.setTag('error', args.exception);
```
serkan-ozal: According to spec, this is the `error` tag definition:
> true if and only if the application considers the operation represented by the Span to have failed

So in case of exception, `error` must be `true`, and there should be error message with `error.message`

PR #25, opened 2018-08-28
[Do Not Merge] Add integrations with mysql, redis, http, pg and aws services

kobalski:


src/Constants.ts
```diff
@@ -37,3 +44,284 @@ export const Syntax = {
...
+export const AwsFirehoseTags = {
+    STREAM_NAME: 'aws.firehose.stream.name',
+};
+
+export const AwsKinesisTags = {
+    STREAM_NAME: 'aws.kinesis.stream.name',
+};
+
+export const AwsLambdaTags = {
+    FUNCTION_NAME: 'function.name',
```
serkan-ozal: They were wrong in the doc, I have updated them with `aws.lambda.` prefix

src/Constants.ts
```diff
@@ -37,3 +44,284 @@ export const Syntax = {
...
+    FUNCTION_QUALIFIER: 'function.qualifier',
+    INVOCATION_TYPE: 'invocation.type',
+    INVOCATION_PAYLOAD: 'invocation.payload',
+};
+
+export const AwsS3Tags = {
+    BUCKET_NAME: 'bucket.name',
```
serkan-ozal: They were wrong in the doc, I have updated them with aws.s3. prefix

src/plugins/Trace.ts
```diff
@@ -140,6 +157,12 @@ export class Trace {
 
         this.tracer.destroy();
         this.instrumenter.unhookModuleCompile();
+
+        for (const key of this.integrations.keys()) {
+            this.integrations.get(key).unwrap();
+        }
+
+        this.integrations = null;
```
serkan-ozal: why we set `integrations` to `null`?

src/plugins/Trace.ts
```diff
@@ -202,6 +228,119 @@ export class Trace {
         return originalEvent;
     }
 
+    private enrichAuditInfo(auditInfo: AuditInfo) {
```
serkan-ozal: I think, these logics should be handled in the integration itself. Otherwise, with every new integration this part will be getting worse

serkan-ozal: You already have `domainName` and `className` as argument to `ThundraSpan`. You can pass them while creating span and then map them as 
- `domainName` of span to `contextGroup` of audit info
- `className` of span to `contextType` of audit info
- `operationName` of span to `contextName` of audit info
- `tags` of span to `props` of audit info
- There might be mismatch between tags and props (such as `operation.type` and `operationType`) with new data model. Therefore, for now you can put both versions of tags for now with a `TODO` comment which says this will be removed with new data model compatible agent version

test/plugins/integrations/integrations.test.js
```diff
@@ -0,0 +1,31 @@
+const Thundra = require('../../../dist/index');
+process.env.AWS_LAMBDA_LOG_STREAM_NAME = '2018/03/02/[$LATEST]applicationId';
+
+describe('Integrations', () => {
```
serkan-ozal: Do we have integration tests? If not, we can remove this test because it doesn't test the features.

kobalski:
Closing this PR, all changes in this PR merged to redesign branch.

PR #33, opened 2018-11-03
[DON'T MERGE] Propagated span from trace entry to exit through entry data ...

serkan-ozal:
... at automated instrumentation


kobalski:
👍

PR #34, opened 2018-11-05
Enable instrumentation hooks before trace plugin initializes

kobalski:
Also integrations are enabled by default now.

### DIFF
--- a/src/plugins/config/TraceConfig.ts
+++ b/src/plugins/config/TraceConfig.ts
@@ -1,10 +1,12 @@
 import BasePluginConfig from './BasePluginConfig';
 import TraceableConfig from './TraceableConfig';
-import {envVariableKeys } from '../../Constants';
+import {envVariableKeys, INTEGRATIONS } from '../../Constants';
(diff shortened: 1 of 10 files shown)


src/plugins/config/TraceConfig.ts
```diff
@@ -42,7 +46,7 @@ class TraceConfig extends BasePluginConfig {
                     ThundraLogger.getInstance().error(`Cannot parse trace def ${key}`);
                 }
             }
-            if (key.startsWith(envVariableKeys.THUNDRA_LAMBDA_TRACE_INTEGRATIONS)) {
+            if (key.startsWith(envVariableKeys.THUNDRA_LAMBDA_TRACE_INTEGRATIONS_DISABLE)) {
```
serkan-ozal: Adding integration even though it is disabled?

serkan-ozal: seems that it is used for disabled integrations, right?

kobalski: yes, I changed the environment variable. It is used for disabling integration.

PR #36, opened 2018-11-14
Add tag support

kobalski:
- Add tag support for invocation support
- Add ability for adding application tags

### DIFF
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -40,7 +40,7 @@ export const envVariableKeys = {
     THUNDRA_AGENT_METRIC_COUNT_AWARE_SAMPLER_COUNT_FREQ: 'thundra_agent_lambda_metric_sample_sampler_countAware_countFreq',
 
     THUNDRA_APPLICATION_TAG_PROP_NAME_PREFIX: 'thundra_agent_lambda_application_tag_',
-    THUNDRA_APPLICATION_ID: 'thundra_agent_application_id',
+    THUNDRA_APPLICATION_ID: 'thundra_agent_lambda_application_id',
     THUNDRA_APPLICATION_NAME: 'thundra_agent_lambda_application_name',
 
 };
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -38,6 +38,11 @@ export const envVariableKeys = {
 
     THUNDRA_AGENT_METRIC_TIME_AWARE_SAMPLER_TIME_FREQ: 'thundra_agent_lambda_metric_sample_sampler_timeAware_timeFreq',
     THUNDRA_AGENT_METRIC_COUNT_AWARE_SAMPLER_COUNT_FREQ: 'thundra_agent_lambda_metric_sample_sampler_countAware_countFreq',
+
+    THUNDRA_APPLICATION_TAG_PROP_NAME_PREFIX: 'thundra_agent_lambda_application_tag_',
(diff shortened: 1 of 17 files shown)


src/Constants.ts
```diff
@@ -38,6 +38,11 @@ export const envVariableKeys = {
 
     THUNDRA_AGENT_METRIC_TIME_AWARE_SAMPLER_TIME_FREQ: 'thundra_agent_lambda_metric_sample_sampler_timeAware_timeFreq',
     THUNDRA_AGENT_METRIC_COUNT_AWARE_SAMPLER_COUNT_FREQ: 'thundra_agent_lambda_metric_sample_sampler_countAware_countFreq',
+
+    THUNDRA_APPLICATION_TAG_PROP_NAME_PREFIX: 'thundra_agent_lambda_application_tag_',
+    THUNDRA_APPLICATION_ID: 'thundra_agent_application_id',
```
serkan-ozal: `thundra_agent_lambda_application_id`?

kobalski: 👍

src/plugins/Utils.ts
```diff
@@ -234,6 +241,28 @@ class Utils {
 
         return monitoringData;
     }
+
+    static addApplicationTags(applicationTags: any): void {
```
serkan-ozal: I think, it is better to get application tags once and reuse them when needed instead of traversing, processing and checking environment variables every time at each invocation

src/plugins/Utils.ts
```diff
@@ -234,6 +241,28 @@ class Utils {
...
+        for (const key of Object.keys(process.env)) {
+            if (key.startsWith(envVariableKeys.THUNDRA_APPLICATION_TAG_PROP_NAME_PREFIX)) {
+                try {
+                    const propsKey = key.substring(envVariableKeys.THUNDRA_APPLICATION_TAG_PROP_NAME_PREFIX.length);
+                    const propsValue = process.env[key];
+                    if (isNaN(parseInt(propsValue, 10))) {
```
serkan-ozal: is it possible to detect and get also decimal numbers (in addition to boolean, integer and string) as application tags?

src/plugins/support/InvocationSupport.ts
```diff
@@ -0,0 +1,48 @@
+import ThundraLogger from '../../ThundraLogger';
+
+class InvocationSupport {
```
serkan-ozal: What about exporting get/set tag methods directly as function in the `InvocationSupport.ts` instead of providing through `InvocationSupport` class. 

I mean, for example, user can just import `setTag` method from `InvocationSupport` js file and use it without `InvocationSupport.getInstance()` since tag related operations are just utility functionalities. @hamitb did similar approach for Python agent

src/plugins/Utils.ts
kobalski: Yes, you are right. I added `ApplicationSupport` class and it parses tags only agent startup.

src/plugins/Utils.ts
kobalski: Ok, right now we can parse float values too. I added test case for float numbers.

src/plugins/support/InvocationSupport.ts
kobalski: Ok. I made all of `InvocationSupport` methods static.  User need not need to call `InvocationSupport.getInstance()` any more. They can directly call methods statically.

PR #38, opened 2018-12-07
Thundra Log Plugin NodeJS console integration

kobalski:
Thundra log plugin automatically shims console methods during initialization. 

Shimmed methods are `log`, `debug`, `info`, `warn`, `error`.

This feature can be disabled via `thundra_agent_lambda_log_console_shim_disable ` environment variable

### DIFF
--- a/src/plugins/Log.ts
+++ b/src/plugins/Log.ts
@@ -4,7 +4,7 @@ import LogData from './data/log/LogData';
 import PluginContext from './PluginContext';
 import MonitoringDataType from './data/base/MonitoringDataType';
 import ThundraTracer from '../opentracing/Tracer';
-import { ConsoleShimmedMethods, logLevels, StdOutLogContext, envVariableKeys, StdErrorLogContext} from '../Constants';
+import { ConsoleShimmedMethods, logLevels, StdOutLogContext, envVariableKeys, StdErrorLogContext } from '../Constants';
 import * as util from 'util';
 
 class Log {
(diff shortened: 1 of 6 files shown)


src/plugins/Log.ts
```diff
@@ -90,14 +95,34 @@ class Log {
...
+        ConsoleShimmedMethods.forEach((method) => {
+            const consoleReference: any = console;
+            const originalConsoleMethod = consoleReference[method].bind(console);
+
+            consoleReference[method] = (...args: any[]) => {
+                const logInfo = {
+                    logMessage:  util.format.apply(util, args),
+                    logLevel: method ? method.toUpperCase() : 'INFO',
+                    logLevelCode: logLevels[method],
+                    logContextName: ConsoleLogContext,
```
serkan-ozal: I think, `ConsoleLogContext` can be `STDOUT` for non-error logs, and `STDERR` for error logs.

src/plugins/Log.ts
```diff
@@ -90,14 +95,34 @@ class Log {
...
+    reportLog(logInfo: any): void {
         const logData = new LogData();
         const activeSpan = this.tracer ? this.tracer.getActiveSpan() : undefined;
         const spanId = activeSpan ? activeSpan.spanContext.spanId : '';
         logData.initWithLogDataValues(this.logData, spanId, logInfo);
 
         this.logs.push(logData);
     }
+
+    shimConsole(): void {
```
serkan-ozal: I think, we need shim console between invocation start and finish. Because otherwise, since there is no active transaction, the logs won't be linked to anywhere. I mean we need to shim at the start of the request and revert back to original at the end of the request

src/plugins/Log.ts
kobalski: Done

src/plugins/Log.ts
kobalski: Done.  Shimming is done at` beforeInvocation` and reverted back at `afterInvocation`

PR #40, opened 2018-12-13
Add Http Trace Propagation and implement extract and inject methods in ThundraTracer

kobalski:


src/opentracing/propagation/TextMap.ts
```diff
@@ -0,0 +1,63 @@
...
+        if (spanContext.baggageItems) {
+            Object.keys(spanContext.baggageItems).forEach((key) => {
+                carrier[TextMapPropagator.BAGGAGE_PREFIX + key] = String(spanContext.baggageItems[key]);
+            });
+        }
+    }
+
+    _extractBaggageItems(carrier: any, spanContext: ThundraSpanContext) {
+        Object.keys(carrier).forEach((key)  => {
+            const pattern = new RegExp(`^${TextMapPropagator.BAGGAGE_PREFIX}(.+)$`);
```
serkan-ozal: might be constant at the `TextMapPropagator` level so there won't be parse for the regexp itself every time

kobalski: Done

src/plugins/integrations/HttpIntegration.ts
```diff
@@ -50,6 +63,16 @@ class HttpIntegration implements Integration {
...
         });
 
+        if (!(Utils.getConfiguration(envVariableKeys.DISABLE_SPAN_CONTEXT_INJECTION) === 'true') ) {
+          const headers = options.headers ? options.headers : {};
+          tracer.inject(span.spanContext, opentracing.FORMAT_TEXT_MAP, headers);
+          options.headers = headers;
+        }
+
+        if (!HttpIntegration.isValidUrl(host)) {
+          return request.apply(this, [options, callback]);
```
serkan-ozal: I think, in case of invalid url that we are not interested in, headers should not be passed in the `options`

kobalski: Done, if the url is invalid we do not create span and inject any headers.

src/plugins/Trace.ts
```diff
@@ -63,20 +65,45 @@ export class Trace {
...
+
+            this.tracer.transactionId = this.pluginContext.transactionId;
+
+            this.rootSpan = this.tracer._startSpan(originalContext.functionName, {
+                rootTraceId: this.pluginContext.traceId,
+                domainName: DomainNames.API,
+                className: ClassNames.LAMBDA,
+            });
+
+            this.rootSpan.spanContext.parentId = propagatedSpanContext.spanId;
```
serkan-ozal: Can't this be done by passing span context to `_startSpan` here? (Passing `traceId`, `transactionId`, `parentId` and `baggageItems` at once) So we don't need to modify span context after created it

kobalski: Done, added `parentContext` option to `_startSpan ` method.

PR #43, opened 2018-12-22
Check topicName and queueName for undefined

kobalski:


src/plugins/integrations/AWSIntegration.ts
```diff
@@ -129,7 +128,7 @@ class AWSIntegration implements Integration {
           } else if (serviceName === 'sns') {
             const operationType = SNSRequestTypes[operationName];
             let topicName = request.params ? request.params.TopicArn : AWS_SERVICE_REQUEST;
-            topicName = topicName.substring(topicName.lastIndexOf(':') + 1);
+            topicName = topicName ? topicName.substring(topicName.lastIndexOf(':') + 1) : topicName;
```
serkan-ozal: When `topicName` is undefined (because there is no ARN in the request params), that undefined value is used as span name below. So as far as I see that might cause another problem because operation name of the span becomes undefined. 

Therefore, in this case, topic name should not be exist in the tags but operation name can be value of `AWS_SERVICE_REQUEST` constant

kobalski: 👍 Made the changes

kobalski: You are right. Closing this PR will send another PR for handling cases which may result spans with undefined operation names.

PR #45, opened 2019-01-15
Fix AWS Integration issues with resource names and operation name being undefined

kobalski:
Also fixes issues below :

- Make the count and time sampling enabled by default.  
- Add support for injecting trigger information when lambda is called API Gateway Passthrough Request.
- Remove extra space character at the end of the `aws.lambda.invocation.request_id` tags.

### DIFF
--- a/src/plugins/config/CountAwareSamplerConfig.ts
+++ b/src/plugins/config/CountAwareSamplerConfig.ts
@@ -1,4 +1,6 @@
 import BasePluginConfig from './BasePluginConfig';
+import Utils from '../utils/Utils';
+import { envVariableKeys } from '../../Constants';
 const koalas = require('koalas');
 
 class CountAwareSamplerConfig extends BasePluginConfig {
@@ -6,8 +8,11 @@ class CountAwareSamplerConfig extends BasePluginConfig {
 
     constructor(options: any) {
         options = options ? options : {};
-        super(koalas(options.enabled, false));
-        this.countFreq = options.countFreq;
+        super(koalas(options.enabled, true));
+
+        const freq = koalas(Utils.getConfiguration(
+            envVariableKeys.THUNDRA_AGENT_METRIC_COUNT_AWARE_SAMPLER_COUNT_FREQ) , options.countFreq, 100);
+        this.countFreq = parseInt(freq, 10);
     }
 }
 
--- a/src/plugins/integrations/AWSIntegration.ts
+++ b/src/plugins/integrations/AWSIntegration.ts
@@ -74,8 +74,10 @@ class AWSIntegration implements Integration {
     function wrapper(wrappedFunction: any) {
       return function AWSSDKWrapper(callback: any) {
 
-        const tracer = ThundraTracer.getInstance();
+        let activeSpan: ThundraSpan;
         try {
+          const tracer = ThundraTracer.getInstance();
+
           const request = this;
           const serviceEndpoint = request.service.config.endpoint;
           const serviceName = Utils.getServiceName(serviceEndpoint as string);
@@ -84,17 +86,15 @@ class AWSIntegration implements Integration {
 
           request.params = request.params ? request.params : {};
 
-          let activeSpan: ThundraSpan = null;
-
           const operationName = request.operation ? request.operation  : AWS_SERVICE_REQUEST;
 
           if (serviceName === 'sqs') {
             const operationType = SQSRequestTypes[operationName];
-            let queueName = koalas(Utils.getQueueName(request.params.QueueUrl), AWS_SERVICE_REQUEST);
+            let queueName = Utils.getQueueName(request.params.QueueUrl);
+            queueName = queueName ? queueName.substring(queueName.lastIndexOf('/') + 1) : queueName;
 
-            queueName = queueName.substring(queueName.lastIndexOf('/') + 1);
-
-            activeSpan = tracer._startSpan(queueName, {
+            const spanName = koalas(queueName, AWS_SERVICE_REQUEST);
+            activeSpan = tracer._startSpan(spanName, {
               childOf: parentSpan,
               domainName: DomainNames.MESSAGING,
               className: ClassNames.SQS,
@@ -102,7 +102,7 @@ class AWSIntegration implements Integration {
               tags: {
                 [SpanTags.SPAN_TYPE]: SpanTypes.AWS_SQS,
                 [AwsSDKTags.REQUEST_NAME]: operationName,
-                [SpanTags.OPERATION_TYPE]: operationType ? operationType : 'READ',
+                [SpanTags.OPERATION_TYPE]: operationType ? operationType : '',
               },
             });
 
@@ -130,11 +130,13 @@ class AWSIntegration implements Integration {
             }
           } else if (serviceName === 'sns') {
             const operationType = SNSRequestTypes[operationName];
-            let topicName = koalas(request.params.TopicArn, AWS_SERVICE_REQUEST);
 
+            let topicName = koalas(request.params.TopicArn, AWS_SERVICE_REQUEST);
             topicName = topicName.substring(topicName.lastIndexOf(':') + 1);
 
-            activeSpan = tracer._startSpan(topicName, {
+            const spanName = koalas(topicName, AWS_SERVICE_REQUEST);
+
+            activeSpan = tracer._startSpan(spanName, {
               childOf: parentSpan,
               domainName: DomainNames.MESSAGING,
               className: ClassNames.SNS,
@@ -142,7 +144,7 @@ class AWSIntegration implements Integration {
               tags: {
                 [SpanTags.SPAN_TYPE]: SpanTypes.AWS_SNS,
                 [AwsSDKTags.REQUEST_NAME]: operationName,
-                [SpanTags.OPERATION_TYPE]: operationType ? operationType : 'READ',
+                [SpanTags.OPERATION_TYPE]: operationType ? operationType : '',
               },
             });
 
@@ -165,9 +167,11 @@ class AWSIntegration implements Integration {
             }
           } else if (serviceName === 'dynamodb') {
             const statementType = DynamoDBRequestTypes[operationName];
-            const tableName = koalas(Utils.getDynamoDBTableName(request), AWS_SERVICE_REQUEST);
+            const tableName = Utils.getDynamoDBTableName(request);
 
-            activeSpan = tracer._startSpan(tableName, {
+            const spanName = koalas(tableName, AWS_SERVICE_REQUEST);
+
+            activeSpan = tracer._startSpan(spanName, {
               childOf: parentSpan,
               domainName: DomainNames.DB,
               className: ClassNames.DYNAMODB,
@@ -175,8 +179,8 @@ class AWSIntegration implements Integration {
               tags: {
                 [DB_TYPE]: DBTypes.DYNAMODB,
                 [DB_INSTANCE]: serviceEndpoint,
-                [DBTags.DB_STATEMENT_TYPE]: statementType ? statementType : 'READ',
-                [SpanTags.OPERATION_TYPE]: statementType ? statementType : 'READ',
+                [DBTags.DB_STATEMENT_TYPE]: statementType ? statementType : '',
+                [SpanTags.OPERATION_TYPE]: statementType ? statementType : '',
                 [SpanTags.SPAN_TYPE]: SpanTypes.AWS_DYNAMO,
                 [AwsSDKTags.REQUEST_NAME]: operationName,
                 [DBTags.DB_STATEMENT]: request.params,
@@ -192,16 +196,16 @@ class AWSIntegration implements Integration {
             }
           } else if (serviceName === 's3') {
             const operationType = S3RequestTypes[operationName];
-            const bucketName = koalas(request.params.Bucket, AWS_SERVICE_REQUEST);
+            const spanName = koalas(request.params.Bucket, AWS_SERVICE_REQUEST);
 
-            activeSpan = tracer._startSpan(bucketName, {
+            activeSpan = tracer._startSpan(spanName, {
               childOf: parentSpan,
               domainName: DomainNames.STORAGE,
               className: ClassNames.S3,
               disableActiveStart: true,
               tags: {
                 [SpanTags.SPAN_TYPE]: SpanTypes.AWS_S3,
-                [SpanTags.OPERATION_TYPE]: operationType ? operationType : 'READ',
+                [SpanTags.OPERATION_TYPE]: operationType ? operationType : '',
                 [AwsS3Tags.BUCKET_NAME]: request.params.Bucket,
                 [AwsSDKTags.REQUEST_NAME]: operationName,
               },
@@ -216,9 +220,9 @@ class AWSIntegration implements Integration {
             }
           } else if (serviceName === 'lambda') {
             const operationType = LambdaRequestType[operationName];
-            const lambdaName = koalas(request.params.FunctionName, AWS_SERVICE_REQUEST);
+            const spanName = koalas(request.params.FunctionName, AWS_SERVICE_REQUEST);
 
-            activeSpan = tracer._startSpan(lambdaName, {
+            activeSpan = tracer._startSpan(spanName, {
               childOf: parentSpan,
               disableActiveStart: true,
               domainName: DomainNames.API,
@@ -242,7 +246,7 @@ class AWSIntegration implements Integration {
               activeSpan.setTag(SpanTags.TRIGGER_DOMAIN_NAME, LAMBDA_APPLICATION_DOMAIN_NAME);
               activeSpan.setTag(SpanTags.TRIGGER_CLASS_NAME, LAMBDA_APPLICATION_CLASS_NAME);
 
-              activeSpan.setTag(AwsLambdaTags.FUNCTION_NAME, lambdaName);
+              activeSpan.setTag(AwsLambdaTags.FUNCTION_NAME, spanName);
             }
 
             if (custom) {
@@ -260,16 +264,16 @@ class AWSIntegration implements Integration {
               }
             }
           } else if (serviceName === 'kinesis') {
-            const streamName = koalas(request.params.StreamName, AWS_SERVICE_REQUEST);
+            const spanName = koalas(request.params.StreamName, AWS_SERVICE_REQUEST);
 
-            activeSpan = tracer._startSpan(streamName, {
+            activeSpan = tracer._startSpan(spanName, {
               childOf: parentSpan,
               domainName: DomainNames.STREAM,
               className: ClassNames.KINESIS,
               disableActiveStart: true,
               tags: {
                 [SpanTags.OPERATION_TYPE]: KinesisRequestTypes[operationName] ?
-                  KinesisRequestTypes[operationName] : 'READ',
+                  KinesisRequestTypes[operationName] : '',
                 [SpanTags.SPAN_TYPE]: SpanTypes.AWS_KINESIS,
                 [AwsSDKTags.REQUEST_NAME]: operationName,
               },
@@ -281,19 +285,19 @@ class AWSIntegration implements Integration {
               activeSpan.setTag(SpanTags.TRIGGER_DOMAIN_NAME, LAMBDA_APPLICATION_DOMAIN_NAME);
               activeSpan.setTag(SpanTags.TRIGGER_CLASS_NAME, LAMBDA_APPLICATION_CLASS_NAME);
 
-              activeSpan.setTag(AwsKinesisTags.STREAM_NAME, streamName);
+              activeSpan.setTag(AwsKinesisTags.STREAM_NAME, request.params.StreamName);
             }
           } else if (serviceName === 'firehose') {
-            const streamName = koalas(request.params.DeliveryStreamName, AWS_SERVICE_REQUEST);
+            const spanName = koalas(request.params.DeliveryStreamName, AWS_SERVICE_REQUEST);
 
-            activeSpan = tracer._startSpan(streamName, {
+            activeSpan = tracer._startSpan(spanName, {
               childOf: parentSpan,
               domainName: DomainNames.STREAM,
               className: ClassNames.FIREHOSE,
               disableActiveStart: true,
               tags: {
                 [SpanTags.OPERATION_TYPE]: FirehoseRequestTypes[operationName] ?
-                  FirehoseRequestTypes[operationName] : 'READ',
+                  FirehoseRequestTypes[operationName] : '',
                 [SpanTags.SPAN_TYPE]: SpanTypes.AWS_FIREHOSE,
                 [AwsSDKTags.REQUEST_NAME]: operationName,
               },
@@ -305,7 +309,7 @@ class AWSIntegration implements Integration {
               activeSpan.setTag(SpanTags.TRIGGER_DOMAIN_NAME, LAMBDA_APPLICATION_DOMAIN_NAME);
               activeSpan.setTag(SpanTags.TRIGGER_CLASS_NAME, LAMBDA_APPLICATION_CLASS_NAME);
 
-              activeSpan.setTag(AwsFirehoseTags.STREAM_NAME, streamName);
+              activeSpan.setTag(AwsFirehoseTags.STREAM_NAME, request.params.DeliveryStreamName);
             }
           } else {
             activeSpan = tracer._startSpan(AWS_SERVICE_REQUEST, {
@@ -337,6 +341,10 @@ class AWSIntegration implements Integration {
           return wrappedFunction.apply(this, [originalCallback]);
 
         } catch (error) {
+          if (activeSpan) {
+            activeSpan.close();
+          }
+
           ThundraLogger.getInstance().error(error);
           return wrappedFunction.apply(this, [callback]);
         }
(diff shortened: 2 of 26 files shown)


src/plugins/config/CountAwareSamplerConfig.ts
```diff
@@ -1,13 +1,18 @@
 import BasePluginConfig from './BasePluginConfig';
+import Utils from '../utils/Utils';
+import { envVariableKeys } from '../../Constants';
 const koalas = require('koalas');
 
 class CountAwareSamplerConfig extends BasePluginConfig {
```
serkan-ozal: Regardsless from `CountAwareSamplerConfig` and `TimeAwareSamplerConfig`, If I talk about `MetricSamplerConfig`, there should be sampling results should be `OR` with eachother (should not override themselves).

For the following code snippet from the `MetricSamplerConfig`:
```
        if (this.countAwareSampler) {
            isSampled = this.countAwareSampler.isSampled();
        }

        if (this.timeAwareSampler) {
            isSampled = this.timeAwareSampler.isSampled();
        }

        if (this.customSampler) {
            if (typeof this.customSampler === 'function' && this.customSampler().isSampled) {
                isSampled = this.customSampler().isSampled();
            }
        }
```

It should something like that:
```
       if (this.countAwareSampler) {
            isSampled = isSampled || this.countAwareSampler.isSampled();
        }

        if (this.timeAwareSampler) {
            isSampled = isSampled || this.timeAwareSampler.isSampled();
        }

        if (this.customSampler) {
            if (typeof this.customSampler === 'function' && this.customSampler().isSampled) {
                isSampled = isSampled || this.customSampler().isSampled();
            }
        }
```

In here, we are calling each sampler even though one  of them returns `true`. The reason is that they might require to be invoked at every check (for ex. count based sampler)

serkan-ozal: Also, I think similar changes should be applied to `TraceSamplerConfig` also

kobalski: I applied the OR logic to  both `MetricSamplerConfig ` and `TraceSamplerConfig `

src/plugins/integrations/RedisIntegration.ts
```diff
@@ -40,66 +42,71 @@ class RedisIntegration implements Integration {
...
+            originalCallback(err, res);
+          };
+
+          options.callback = wrappedCallback;
 
-        return internalSendCommand.call(this, options);
+          return internalSendCommand.call(this, options);
+        } catch (error) {
+          ThundraLogger.getInstance().error(error);
+          internalSendCommand.call(this, options);
```
serkan-ozal: In here, if `internalSendCommand` command itself fails, do we send another call to Redis?

src/plugins/integrations/PostgreIntegration.ts
```diff
@@ -41,66 +43,69 @@ class PostgreIntegration implements Integration {
...
 
+            if (originalCallback) {
+              originalCallback(err, res);
+            }
+          };
+
+          return pgQuery;
+        } catch (error) {
+          ThundraLogger.getInstance().error(error);
+          query.apply(this, arguments);
```
serkan-ozal: Can we be sure that in case of error `query.apply` is not called twice

src/plugins/integrations/PostgreIntegration.ts
```diff
@@ -41,66 +43,69 @@ class PostgreIntegration implements Integration {
...
-            const parseError = Utils.parseError(err);
-            span.setTag('error', true);
-            span.setTag('error.kind', parseError.errorType);
-            span.setTag('error.message', parseError.errorMessage);
+          if (statement) {
+            const statementType = statement.split(' ')[0].toUpperCase();
+            span.addTags({
+              [DBTags.DB_STATEMENT_TYPE]: statementType,
+              [DBTags.DB_STATEMENT]: statement,
+              [SpanTags.OPERATION_TYPE]: SQLQueryOperationTypes[statementType] ? SQLQueryOperationTypes[statementType] : 'READ',
```
serkan-ozal: we might return empty operation type instead of `READ` if the operation is not known

src/plugins/integrations/MySQL2Integration.ts
```diff
@@ -40,68 +42,73 @@ class MySQL2Integration implements Integration {
...
+          } else {
+            sequence.on('end', () => {
+              span.close();
+            });
+          }
+
+          return sequence;
+        } catch (error) {
+          ThundraLogger.getInstance().error(error);
+          query.call(this, sql, values, cb);
```
serkan-ozal: Can we be sure that in case of error `query.call` is not called twice

src/plugins/integrations/MySQL2Integration.ts
```diff
@@ -40,68 +42,73 @@ class MySQL2Integration implements Integration {
...
-            span.setTag('error.stack', parseError.stack);
-            span.setTag('error.code', parseError.code);
+          const statement = sequence.sql;
+
+          if (statement) {
+            const statementType = statement.split(' ')[0].toUpperCase();
+            span.addTags({
+              [DBTags.DB_STATEMENT_TYPE]: statementType,
+              [DBTags.DB_STATEMENT]: statement,
+              [SpanTags.OPERATION_TYPE]: SQLQueryOperationTypes[statementType] ? SQLQueryOperationTypes[statementType] : 'READ',
```
serkan-ozal: we might return empty operation type instead of `READ` if the operation is not known

src/plugins/integrations/HttpIntegration.ts
```diff
@@ -48,73 +51,77 @@ class HttpIntegration implements Integration {
...
+            span.setTag('error.stack', parseError.stack);
+            span.setTag('error.code', parseError.code);
+
+            span.close();
+          });
+
+          return req;
+        } catch (error) {
+          ThundraLogger.getInstance().error(error);
           return request.apply(this, [options, callback]);
```
serkan-ozal: Can we be sure that in case of error `request.apply` is not called twice

src/plugins/integrations/HttpIntegration.ts
```diff
@@ -48,73 +51,77 @@ class HttpIntegration implements Integration {
...
+
+          if (!(Utils.getConfiguration(envVariableKeys.DISABLE_SPAN_CONTEXT_INJECTION) === 'true')) {
+            const headers = options.headers ? options.headers : {};
+            tracer.inject(span.spanContext, opentracing.FORMAT_TEXT_MAP, headers);
+            options.headers = headers;
+          }
 
-        if (!HttpIntegration.isValidUrl(host)) {
+          span.addTags({
+            [SpanTags.OPERATION_TYPE]: 'CALL',
```
serkan-ozal: I think, anymore we can set operation type to the `method` here instead of `CALL` constant.

src/plugins/integrations/MySQL2Integration.ts
```diff
@@ -40,68 +42,73 @@ class MySQL2Integration implements Integration {
...
-        if (statement) {
-          const statementType = statement.split(' ')[0].toUpperCase();
+        try {
+          const tracer = ThundraTracer.getInstance();
+          const parentSpan = tracer.getActiveSpan();
+
+          const span = tracer._startSpan(this.config.database, {
+            childOf: parentSpan,
+            domainName: DomainNames.DB,
+            className: ClassNames.RDB,
```
serkan-ozal: I think, anymore we can set class name to the `DBTypes.MYSQL` here instead of `RDB` constant.

src/plugins/integrations/PostgreIntegration.ts
```diff
@@ -41,66 +43,69 @@ class PostgreIntegration implements Integration {
...
-            [SpanTags.TRIGGER_OPERATION_NAMES]: [tracer.functionName],
+        try {
+          const tracer = ThundraTracer.getInstance();
+          const parentSpan = tracer.getActiveSpan();
+
+          const params = this.connectionParameters;
+          const span = tracer._startSpan(params.database, {
+            childOf: parentSpan,
+            domainName: DomainNames.DB,
+            className: ClassNames.RDB,
```
serkan-ozal: I think, anymore we can set class name to the `DBTypes. PG` here instead of `RDB` constant.

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -82,14 +82,15 @@ class AWSIntegration implements Integration {
...
+
           let activeSpan: ThundraSpan = null;
 
           const operationName = request.operation ? request.operation  : AWS_SERVICE_REQUEST;
 
           if (serviceName === 'sqs') {
             const operationType = SQSRequestTypes[operationName];
-            let queueName = operationType ?
-            Utils.getQueueName(request.params.QueueUrl) : AWS_SERVICE_REQUEST;
+            let queueName = koalas(Utils.getQueueName(request.params.QueueUrl), AWS_SERVICE_REQUEST);
```
serkan-ozal: If queue name is not valid for the request, it is ok to use `AWS_SERVICE_REQUEST ` as the default span name, but it should not be used for `AwsSQSTags.QUEUE_NAME` tag. So you may handle this by introducing another local variable for span name.

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -128,7 +130,8 @@ class AWSIntegration implements Integration {
             }
           } else if (serviceName === 'sns') {
             const operationType = SNSRequestTypes[operationName];
-            let topicName = request.params ? request.params.TopicArn : AWS_SERVICE_REQUEST;
+            let topicName = koalas(request.params.TopicArn, AWS_SERVICE_REQUEST);
```
serkan-ozal: If topic name is not valid for the request, it is ok to use `AWS_SERVICE_REQUEST ` as the default span name, but it should not be used for `AwsSNSTags.QUEUE_NAME` tag. So you may handle this by introducing another local variable for span name.

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -161,7 +165,7 @@ class AWSIntegration implements Integration {
             }
           } else if (serviceName === 'dynamodb') {
             const statementType = DynamoDBRequestTypes[operationName];
-            const tableName = statementType ? Utils.getDynamoDBTableName(request) : AWS_SERVICE_REQUEST;
+            const tableName = koalas(Utils.getDynamoDBTableName(request), AWS_SERVICE_REQUEST);
```
serkan-ozal: If table name is not valid for the request, it is ok to use `AWS_SERVICE_REQUEST ` as the default span name, but it should not be used for `AwsDynamoTags.TABLE_NAME` tag. So you may handle this by introducing another local variable for span name.

serkan-ozal: Additionally, as far as I see, currently `Utils.getDynamoDBTableName(request)` always return a non empty value

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -209,10 +212,11 @@ class AWSIntegration implements Integration {
               activeSpan.setTag(SpanTags.TOPOLOGY_VERTEX, true);
               activeSpan.setTag(SpanTags.TRIGGER_DOMAIN_NAME, LAMBDA_APPLICATION_DOMAIN_NAME);
               activeSpan.setTag(SpanTags.TRIGGER_CLASS_NAME, LAMBDA_APPLICATION_CLASS_NAME);
+              activeSpan.setTag(AwsS3Tags.OBJECT_NAME, request.params.Key);
             }
           } else if (serviceName === 'lambda') {
             const operationType = LambdaRequestType[operationName];
-            const lambdaName = operationType ? request.params.FunctionName : AWS_SERVICE_REQUEST;
+            const lambdaName = koalas(request.params.FunctionName, AWS_SERVICE_REQUEST);
```
serkan-ozal: If function name is not valid for the request, it is ok to use `AWS_SERVICE_REQUEST` as the default span name, but it should not be used for `AwsLambdaTags. FUNCTION_NAME` tag. So you may handle this by introducing another local variable for span name.

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -255,7 +260,7 @@ class AWSIntegration implements Integration {
               }
             }
           } else if (serviceName === 'kinesis') {
-            const streamName = request.params ? request.params.StreamName : AWS_SERVICE_REQUEST;
+            const streamName = koalas(request.params.StreamName, AWS_SERVICE_REQUEST);
```
serkan-ozal: If stream name is not valid for the request, it is ok to use `AWS_SERVICE_REQUEST` as the default span name, but it should not be used for `AwsKinesisTags. STREAM_NAME` tag. So you may handle this by introducing another local variable for span name

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -276,9 +280,12 @@ class AWSIntegration implements Integration {
               activeSpan.setTag(SpanTags.TOPOLOGY_VERTEX, true);
               activeSpan.setTag(SpanTags.TRIGGER_DOMAIN_NAME, LAMBDA_APPLICATION_DOMAIN_NAME);
               activeSpan.setTag(SpanTags.TRIGGER_CLASS_NAME, LAMBDA_APPLICATION_CLASS_NAME);
+
+              activeSpan.setTag(AwsKinesisTags.STREAM_NAME, streamName);
             }
           } else if (serviceName === 'firehose') {
-            const streamName = request.params ? request.params.DeliveryStreamName : AWS_SERVICE_REQUEST;
+            const streamName = koalas(request.params.DeliveryStreamName, AWS_SERVICE_REQUEST);
```
serkan-ozal: If stream name is not valid for the request, it is ok to use `AWS_SERVICE_REQUEST` as the default span name, but it should not be used for `AwsFirehoseTags. STREAM_NAME` tag. So you may handle this by introducing another local variable for span name

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -330,6 +338,7 @@ class AWSIntegration implements Integration {
 
         } catch (error) {
           ThundraLogger.getInstance().error(error);
+          return wrappedFunction.apply(this, [callback]);
```
serkan-ozal: Can we be sure that in case of error `wrappedFunction.apply` is not called twice

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -209,10 +212,11 @@ class AWSIntegration implements Integration {
               activeSpan.setTag(SpanTags.TOPOLOGY_VERTEX, true);
               activeSpan.setTag(SpanTags.TRIGGER_DOMAIN_NAME, LAMBDA_APPLICATION_DOMAIN_NAME);
               activeSpan.setTag(SpanTags.TRIGGER_CLASS_NAME, LAMBDA_APPLICATION_CLASS_NAME);
+              activeSpan.setTag(AwsS3Tags.OBJECT_NAME, request.params.Key);
```
serkan-ozal: What if `request.params.Key` is undefined? I mean, there is no `Key` in the request

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -185,10 +188,11 @@ class AWSIntegration implements Integration {
               activeSpan.setTag(SpanTags.TOPOLOGY_VERTEX, true);
               activeSpan.setTag(SpanTags.TRIGGER_DOMAIN_NAME, LAMBDA_APPLICATION_DOMAIN_NAME);
               activeSpan.setTag(SpanTags.TRIGGER_CLASS_NAME, LAMBDA_APPLICATION_CLASS_NAME);
+              activeSpan.setTag(AwsDynamoTags.TABLE_NAME, tableName);
             }
           } else if (serviceName === 's3') {
             const operationType = S3RequestTypes[operationName];
-            const bucketName = operationType ? request.params.Bucket : AWS_SERVICE_REQUEST;
+            const bucketName = koalas(request.params.Bucket, AWS_SERVICE_REQUEST);
```
serkan-ozal: If bucket name is not valid for the request, it is ok to use `AWS_SERVICE_REQUEST` as the default span name, but it should not be used for `AwsS3Tags. OBJECT_NAME` tag. So you may handle this by introducing another local variable for span name.

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -200,7 +204,6 @@ class AWSIntegration implements Integration {
                 [SpanTags.OPERATION_TYPE]: operationType ? operationType : 'READ',
                 [AwsS3Tags.BUCKET_NAME]: request.params.Bucket,
```
serkan-ozal: What if `request.params.Bucket` is undefined? I mean, there is no `Bucket` in the request

src/plugins/integrations/RedisIntegration.ts
```diff
@@ -40,66 +42,71 @@ class RedisIntegration implements Integration {
...
-          originalCallback(err, res);
-        };
+          if (this.connection_options) {
+            host = String(this.connection_options.host);
+            port = String(this.connection_options.port);
+            command = options.command.toUpperCase();
+          }
 
-        options.callback = wrappedCallback;
+          const operationType = RedisCommandTypes[command] ? RedisCommandTypes[command] : 'READ';
```
serkan-ozal: we might use empty operation type instead of `READ` if the operation is not known

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -289,7 +296,6 @@ class AWSIntegration implements Integration {
                   FirehoseRequestTypes[operationName] : 'READ',
```
serkan-ozal: we might use empty operation type instead of `READ` if the operation is not known

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -267,7 +272,6 @@ class AWSIntegration implements Integration {
                   KinesisRequestTypes[operationName] : 'READ',
```
serkan-ozal: we might use empty operation type instead of `READ` if the operation is not known

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -200,7 +204,6 @@ class AWSIntegration implements Integration {
                 [SpanTags.OPERATION_TYPE]: operationType ? operationType : 'READ',
```
serkan-ozal: we might use empty operation type instead of `READ` if the operation is not known

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -174,7 +178,6 @@ class AWSIntegration implements Integration {
                 [DBTags.DB_STATEMENT_TYPE]: statementType ? statementType : 'READ',
```
serkan-ozal: we might use empty statement type instead of `READ` if the statement is not known

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -174,7 +178,6 @@ class AWSIntegration implements Integration {
...
                 [SpanTags.OPERATION_TYPE]: statementType ? statementType : 'READ',
```
serkan-ozal: we might use empty operation type instead of `READ` if the operation is not known

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -139,7 +142,6 @@ class AWSIntegration implements Integration {
               tags: {
                 [SpanTags.SPAN_TYPE]: SpanTypes.AWS_SNS,
                 [AwsSDKTags.REQUEST_NAME]: operationName,
-                [AwsSNSTags.TOPIC_NAME]: topicName,
                 [SpanTags.OPERATION_TYPE]: operationType ? operationType : 'READ',
```
serkan-ozal: we might use empty operation type instead of `READ` if the operation is not known

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -101,7 +102,6 @@ class AWSIntegration implements Integration {
               tags: {
                 [SpanTags.SPAN_TYPE]: SpanTypes.AWS_SQS,
                 [AwsSDKTags.REQUEST_NAME]: operationName,
-                [AwsSQSTags.QUEUE_NAME]: queueName,
                 [SpanTags.OPERATION_TYPE]: operationType ? operationType : 'READ',
```
serkan-ozal: we might use empty operation type instead of `READ` if the operation is not known

src/plugins/utils/Utils.ts
```diff
@@ -167,17 +167,20 @@ class Utils {
...
+        if (request.params && request.params.TableName) {
             tableName = request.params.TableName;
         }
-        if (request.params.RequestItems) {
+
+        if (request.params && request.params.RequestItems) {
             tableName = Object.keys(request.params.RequestItems).join(',');
         }
+
         return tableName;
```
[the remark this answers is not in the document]
kobalski: Done.

src/plugins/integrations/RedisIntegration.ts
kobalski: In case of error in the wrapped function error callback is called. I verified, it is not called twice.

src/plugins/integrations/PostgreIntegration.ts
kobalski: In case of error in the wrapped function error callback is called. I verified, it is not called twice.

src/plugins/integrations/PostgreIntegration.ts
kobalski: Done.

src/plugins/integrations/MySQL2Integration.ts
kobalski: In case of error in the wrapped function error callback is called. I verified, it is not called twice.

src/plugins/integrations/MySQL2Integration.ts
kobalski: Done

src/plugins/integrations/HttpIntegration.ts
kobalski: In case of error in the wrapped function error callback is called. I verified, it is not called twice.

src/plugins/integrations/HttpIntegration.ts
kobalski: Done.

src/plugins/integrations/MySQL2Integration.ts
kobalski: Done. Classname is set `MYSQL` now.

src/plugins/integrations/AWSIntegration.ts
kobalski: Done. Introduced a `spanName` variable.

src/plugins/integrations/AWSIntegration.ts
kobalski: Done. Introduced a `spanName` variable.

src/plugins/integrations/AWSIntegration.ts
kobalski: Done. Introduced a `spanName` variable. `Utils.getDynamoDBTableName(request)` returns null now.

src/plugins/integrations/AWSIntegration.ts
kobalski: Done. Introduced a `spanName` variable.

src/plugins/integrations/AWSIntegration.ts
kobalski: Done. Introduced a `spanName` variable.

src/plugins/integrations/AWSIntegration.ts
kobalski: In case of error in the wrapped function error callback is called. I verified, it is not called twice.

src/plugins/integrations/AWSIntegration.ts
kobalski: Then `AwsS3Tags.OBJECT_NAME` will be `undefined` and it will not be serialized during reporting.

src/plugins/integrations/AWSIntegration.ts
kobalski: Done. Introduced a spanName variable.

src/plugins/integrations/AWSIntegration.ts
kobalski: Then `AwsS3Tags.BUCKET_NAME` will be undefined and it will not be serialized during reporting.

src/plugins/integrations/RedisIntegration.ts
kobalski: Done

src/plugins/integrations/AWSIntegration.ts
kobalski: Done

src/plugins/integrations/AWSIntegration.ts
kobalski: Done

src/plugins/integrations/AWSIntegration.ts
kobalski: Done

src/plugins/integrations/AWSIntegration.ts
kobalski: Done

src/plugins/integrations/AWSIntegration.ts
kobalski: Done

src/plugins/integrations/AWSIntegration.ts
kobalski: Done

src/plugins/integrations/AWSIntegration.ts
kobalski: Done

Issue #54, opened 2019-04-11
`UnhandledPromiseRejectionWarning` message when throwing error in async lambda invocation


kobalski:
Below code  logs `UnhandledPromiseRejectionWarning`.

```
const thundra = require('@thundra/core')();
exports.handler = thundra(async (event) => {
    throw new Error('test error');
});
```

Issue #52, opened 2019-03-28
aws-sdk does not close span when used with promise instead of callbacks


kobalski:
The below code creates a span which does not finish even though the `putObjectPromise` is resolved.

```
var s3 = new AWS.S3({apiVersion: '2006-03-01', region: 'us-west-2'});
var params = {
  Bucket: 'bucket',
  Key: 'example2.txt',
  Body: 'Uploaded text using the promise-based method!'
};

var putObjectPromise = s3.putObject(params).promise();

putObjectPromise.then(function(data) {
  console.log('Success');
}).catch(function(err) {
  console.log(err);
});
```

PR #57, opened 2019-05-14
Improvements and issue fixes

kobalski:
- Fixes https://github.com/thundra-io/thundra-lambda-agent-nodejs/issues/54
- Fixes https://github.com/thundra-io/thundra-lambda-agent-nodejs/issues/52 
- Fixes `Log Level filtering does not work with console.log integration`
- Fixes `Send data of SNS, SQS, Lambda and HTTP operations with masking capability`
- Fixes ` Don't send Kinesis, Firehose and CloudWatch log event data`
- Fixes  `Nodejs agent composite data model support over CW logs`
- Fixes  ` error.message tag must be string type`
- Fixes `Do not send trace data with trace plugin`


src/plugins/integrations/HttpIntegration.ts
```diff
@@ -115,16 +115,30 @@ class HttpIntegration implements Integration {
...
+          req.emit = function (eventName: any, arg: any) {
+            if (eventName === 'socket') {
+              if (req.listenerCount('response') === 1) {
+                req.on('response', (res: any) => res.resume());
+              }
+
+              if (!config.maskHttpBody && arg._httpMessage._hasBody) {
+                 try {
+                  const lines = arg._httpMessage.output[0].split('\n');
+                  span.setTag(HttpTags.BODY, lines[lines.length - 1]);
```
serkan-ozal: Can't we get request body from `options`?
I mean `const body = options.body || ''`

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -529,23 +546,39 @@ class AWSIntegration implements Integration {
...
+                            }
+                            if (activeSpan) {
+                                activeSpan.closeWithCallback(me, originalCallback, [err, data]);
+                            }
+                        };
 
+                        return originalFunction.apply(this, [wrappedCallback]);
+                    } else {
+                        request.on('error', (error: any) => {
+                            if (error && activeSpan) {
```
serkan-ozal: Don't we close span on error?

kobalski: 👍 Good catch. We should close the span in case of http error. Fixed.

src/plugins/integrations/HttpIntegration.ts
kobalski: Nope. See the docs for http request options docs :  https://nodejs.org/api/http.html#http_http_request_options_callback

Http request options do not have `body` parameter, below code is a typical http post request.
As you can see you write the post data to request object, so I can only extract the post data by listening `socket` event.  `options.body ` does not work. This is the cleanest way, I came up with to request body.

```
const postData =JSON.stringify({
  'msg': 'Hello World!'
});

const options = {
  hostname: 'www.google.com',
  port: 80,
  path: '/upload',
  method: 'POST',
  headers: {
            'Authorization': `Token ${TOKEN}`,
            'Content-Type': 'application/json'
      }
};

const req = http.request(options, (res) => {
});

// Write data to request body
req.write(postData);
req.end();
```

serkan-ozal: How it behaves if body is multi-line? As far as I see, we get the latest line

PR #61, opened 2019-05-17
Add time, count and error based handling to all the plugins.

kobalski:
Also adds support for full configuration for sampling over environment variables.


serkan-ozal:
I think, we need to redesign sampler API and change the way how it is handled:

- I see that in the sampler configs, there are all the samplers are defined and each of them is checked one by one. With the every sampler, this is going to be more complex. Therefore, we should have `CompositeSampler` which gets list of samplers and an operation (`AND` and `OR`) to combine results of each given samplers.

- Currently, sampler configs make sampling decision but they should just be config and should not have sampling behaviour.

- I think, we don't need sampler configs. Samplers to be used can be directly passed in the `InvocationConfig`, `TraceConfig`, `MetricConfig` and `LogConfig` programmatically. No need to make everything through environment variable, especially it will be very complex when we introduce `CompositeSampler`. 

- Configs (`InvocationConfig`, `TraceConfig`, etc ...) can only have one sampler instance which can be `CountAwareSampler`, `TimeAwareSampler`, `CompositeSampler` as a combination of any other samplers (for ex. combination of `CountAwareSampler` and `TimeAwareSampler`) or any custom sampler. So no need to have all the types of samplers in the config.

README.md
```diff
@@ -70,8 +70,8 @@ Check out the [configuration part](https://thundra.readme.io/docs/nodejs-configu
 | thundra_agent_lambda_publish_report_rest_trustAllCertificates            | bool   |    false      |
 | thundra_agent_lambda_debug_enable                                        | bool   |    false      |
 | thundra_agent_lambda_trace_instrument_integrations_disable               | array  |    []         |
-| thundra_agent_lambda_metric_sample_sampler_timeAware_timeFreq            | number |    300000     |
-| thundra_agent_lambda_metric_sample_sampler_countAware_countFreq          | number |    10         |
+| thundra_agent_lambda_sample_sampler_timeAware_timeFreq                   | number |    300000     |
```
serkan-ozal: To be consistent with Java and keep env. variable name shorter, `thundra_agent_lambda_sampler_timeAware_timeFreq` is better

README.md
```diff
@@ -70,8 +70,8 @@ Check out the [configuration part](https://thundra.readme.io/docs/nodejs-configu
...
+| thundra_agent_lambda_sample_sampler_countAware_countFreq                 | number |    10         |
```
serkan-ozal: To be consistent with Java and keep env. variable name shorter, `thundra_agent_lambda_sampler_countAware_countFreq` is better

src/Constants.ts
```diff
@@ -72,6 +69,9 @@ export const envVariableKeys = {
     THUNDRA_MASK_LAMBDA_PAYLOAD: 'thundra_agent_lambda_trace_integrations_aws_lambda_payload_mask',
     THUNDRA_MASK_HTTP_BODY: 'thundra_agent_lambda_trace_integrations_aws_http_body_mask',
     THUNDRA_MASK_ERROR_STACK_TRACE: 'thundra_agent_lambda_error_stacktrace_mask',
+
+    THUNDRA_AGENT_TIME_AWARE_SAMPLER_TIME_FREQ: 'thundra_agent_lambda_sample_sampler_timeAware_timeFreq',
```
serkan-ozal: To be consistent with Java and keep env. variable name shorter, `thundra_agent_lambda_sampler_timeAware_timeFreq` is better

src/Constants.ts
```diff
@@ -72,6 +69,9 @@ export const envVariableKeys = {
...
+    THUNDRA_AGENT_COUNT_AWARE_SAMPLER_COUNT_FREQ: 'thundra_agent_lambda_sample_sampler_countAware_countFreq',
```
serkan-ozal: To be consistent with Java and keep env. variable name shorter, `thundra_agent_lambda_sampler_countAware_countFreq` is better

src/ThundraWrapper.ts
```diff
@@ -152,9 +150,13 @@ class ThundraWrapper {
     }
 
     async executeAfteInvocationAndReport(afterInvocationData: any) {
+        afterInvocationData.error ? InvocationSupport.setErrorenous(true) : InvocationSupport.setErrorenous(false);
+
         await this.executeHook('after-invocation', afterInvocationData, true);
         this.resetTime();
         await this.reporter.sendReports();
+
+        InvocationSupport.setErrorenous(false);
```
serkan-ozal: To make it more safe, can we also have `InvocationSupport.setErrorenous(false);` at the beginning of `invoke` before executing `before-invocation` hooks

src/opentracing/sampler/CompositeSampler.ts
```diff
@@ -0,0 +1,34 @@
...
+    constructor(samplers: Array<Sampler<any>>, operator: SamplerCompositionOperator) {
+        this.samplers =  samplers;
+        this.operator = operator === undefined ? SamplerCompositionOperator.OR : operator;
+    }
+
+    isSampled(arg?: any): boolean {
+        if (this.operator === SamplerCompositionOperator.AND) {
+            let isSampled: boolean = true;
+            for (const sampler of this.samplers) {
+                isSampled = isSampled && sampler.isSampled(arg);
```
serkan-ozal: `sampler.isSampled(arg)` should be at the left side of `&&` operator. Otherwise if `sampled` becomes `false`, right side is not evaluated. And this can be problematic for some samplers like `CountAwareSampler` as it should always be executed to increase the counter.

src/opentracing/sampler/CompositeSampler.ts
```diff
@@ -0,0 +1,34 @@
...
+            }
+            return isSampled;
+        } else {
+            let isSampled: boolean = true;
+            for (const sampler of this.samplers) {
+                isSampled = isSampled || sampler.isSampled(arg);
```
serkan-ozal: `sampler.isSampled(arg)` should be at the left side of `||` operator. Otherwise if `sampled` becomes `true`, right side is not evaluated. And this can be problematic for some samplers like `CountAwareSampler` as it should always be executed to increase the counter.

kobalski:
You can find the examples of the new API in here : https://github.com/thundra-io/thundra-examples-lambda-nodejs/blob/master/sampling-demo/index.js

serkan-ozal:
There are fails in the CircleCI: https://circleci.com/gh/thundra-io/thundra-lambda-agent-nodejs/406

PR #66, opened 2019-05-29
Centralize log filtering and disable log filtering if lambda fails

kobalski:


serkan-ozal:
There is fail for `/metric.test.js`

.circleci/config.yml
```diff
@@ -4,26 +4,6 @@ jobs:
       docker:
         - image: circleci/node:8.11
         
-        - image: postgres:9.5-alpine
```
serkan-ozal: Why we are removing all the integration tests. They can fail on CI until we fix it.

config.json
```diff
@@ -5,6 +5,6 @@
   "collectCoverageFrom": ["dist/*.js", "dist/opentracing/*.js", "dist/opentracing/instrument/*.js", "dist/opentracing/sampler/*.js", "dist/plugins/*.js", 
     "dist/plugins/data/base/*.js", "dist/plugins/data/invocation/*.js", "dist/plugins/data/metric/*.js",
     "dist/plugins/config/*.js", "dist/plugins/data/trace/*.js", "dist/plugins/error/*.js",
-    "dist/plugins/listeners/*.js", "dist/plugins/support/*.js", "index.js", "dist/plugins/integrations/*.js"
+    "dist/plugins/listeners/*.js", "dist/plugins/support/*.js", "index.js"
```
serkan-ozal: Integration tests should not be deleted as they can fail on CI until we fix it

kobalski: Disabled only ES Integration test and removed es image for now.

PR #73, opened 2019-06-21
Enable ES tests and reduce the resource size of circle ci build

kobalski:


kobalski:
Merging this pr it only changed test config.

PR #87, opened 2019-07-19
Handle `ECONNRESET` error when keepAlive connection is terminated by …

kobalski:
…the server. In a long-running lambda container keep-alive connections are reset by our collector servers due to nginx config for `keepalive_timeout` is 30 seconds. If a request fails due to this error, we do not log the error but we try to send the data again otherwise we miss monitoring data.

### DIFF
--- a/src/Reporter.ts
+++ b/src/Reporter.ts
@@ -81,7 +81,7 @@ class Reporter {
                     break;
                 }
 
-                batch.push(Utils.stripCommonFields(report.data  as BaseMonitoringData));
+                batch.push(Utils.stripCommonFields(report.data as BaseMonitoringData));
             }
 
             compositeData.allMonitoringData = batch;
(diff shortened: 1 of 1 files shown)


src/Reporter.ts
```diff
@@ -133,7 +133,13 @@ class Reporter {
             }
         });
 
-        await Promise.all(reportPromises).catch((err) => {
+        await Promise.all(reportPromises).catch(async (err) => {
+            if (err.code === 'ECONNRESET') {
+                ThundraLogger.getInstance().debug(
+                    'Keep Alive connection reset by server. Will send monitoring data again.');
+                await this.sendReports();
```
serkan-ozal: It is better to try only once. Not sure whether it is possible in practice, but in theory, this might result infinite recursion by calling `sendReports` recursively. To prevent this:

```javascript
async sendReports(dontRetry: boolean): Promise<void> {
    ...
    await Promise.all(reportPromises).catch(async (err) => {
            if (dontRetry != true) {
                if (err.code === 'ECONNRESET') {
                    ThundraLogger.getInstance().debug('Connection reset by server. Will send monitoring data again.');
                    await this.sendReports(true);
                    return;
                }
            }
            ThundraLogger.getInstance().error(err);
        });
}
```

kobalski: I included `connectionRetryCount` to Reporter to make sure we try to reconnect to server only once.

PR #88, opened 2019-07-24
Split query params from http operation names

hamitb:


src/plugins/integrations/HttpIntegration.ts
```diff
@@ -102,8 +105,8 @@ class HttpIntegration implements Integration {
                         [SpanTags.SPAN_TYPE]: SpanTypes.HTTP,
                         [HttpTags.HTTP_METHOD]: method,
                         [HttpTags.HTTP_HOST]: host,
-                        [HttpTags.HTTP_PATH]: path.split('?')[0],
-                        [HttpTags.HTTP_URL]: host + path,
+                        [HttpTags.HTTP_PATH]: path,
+                        [HttpTags.HTTP_URL]: `${host}${path}?${queryParams}`,
```
serkan-ozal: When there is no query params, doesn't this add extra/redundant `?` character at the end of the HTTP url?

hamitb: Sure, fixed 👍

PR #95, opened 2019-08-07
Fix/span http error

fatihaydilek:


test/integration/http.integration.test.js
```diff
@@ -35,6 +35,74 @@ describe('HTTP integration', () => {
...
+            httpPathDepth: 2,
+        });
+        const sdk = require('http');
+
+        const tracer = new ThundraTracer();
+        InvocationSupport.setFunctionName('functionName');
+
+        return Http.getError(sdk).then(() => {
+            const span = tracer.getRecorder().spanList[0];
+            console.log(span);
```
hamitb: This line causes unnecessary log on the test output, we can remove it 👍

PR #98, opened 2019-08-16
add resourceMaxDuration to resource

fatihaydilek:


src/plugins/data/invocation/Resource.ts
```diff
@@ -50,6 +53,9 @@ class Resource {
                 });
             }
             this.resourceDuration += resource.resourceDuration;
+            if (resource.resourceDuration > this.resourceMaxDuration) {
```
serkan-ozal: I think this should be `resource.resourceMaxDuration > this.resourceMaxDuration` as we are merging two resources

PR #102, opened 2019-08-30
add resourceAvgDuration to resource

fatihaydilek:


src/plugins/data/invocation/Resource.ts
```diff
@@ -53,12 +56,19 @@ class Resource {
                 });
             }
             this.resourceDuration += resource.resourceDuration;
+            this.updateResourceAvgDuration(resource.resourceDuration);
```
serkan-ozal: Can't this be 
```
this.resourceAvgDuration = this.resourceDuration / this.resourceCount;
```
simply?

PR #106, opened 2019-09-24
fix timeout issue due to unfinished callbacks

fatihaydilek:


src/opentracing/Recorder.ts
```diff
@@ -34,15 +34,13 @@ class ThundraRecorder {
...
-                if (typeof(options.callback) === 'function') {
-                    options.callback.apply(options.me, options.args);
+                if (!isCallbackCalled) {
+                    if (typeof(options.callback) === 'function') {
+                        options.callback.apply(options.me, options.args);
+                    }
                 }
             }
+
         } else if (spanEvent === SpanEvent.SPAN_FINISH) {
```
hamitb: We should do the above changes for the cases where `spanEvent === SpanEvent.SPAN_FINISH` too.

src/opentracing/Recorder.ts
```diff
@@ -34,15 +34,13 @@ class ThundraRecorder {
...
-                if (shouldInvokeCallback) {
-                    shouldInvokeCallback = !isCallbackCalled;
-                }
-            }
-            if (shouldInvokeCallback) {
-                if (typeof(options.callback) === 'function') {
-                    options.callback.apply(options.me, options.args);
+                if (!isCallbackCalled) {
+                    if (typeof(options.callback) === 'function') {
+                        options.callback.apply(options.me, options.args);
```
hamitb: Is it correct to call the callback in the `SpanEvent.SPAN_START` event?

PR #130, opened 2020-01-03
Online Debugging

hamitb:


src/debugBridge.ts
```diff
@@ -0,0 +1,100 @@
...
+        process.send(err.message);
+    }
+    log('brokerSocket:', err.message);
+});
+
+process.on('SIGTERM', () => {
+    if (brokerSocket.readyState === WebSocket.OPEN) {
+        brokerSocket.close(CLOSING_CODES.NORMAL, 'Normal');
+    }
+    if (debuggerSocket.destroyed) {
```
serkan-ozal: `if (!debuggerSocket.destroyed)`

src/ThundraWrapper.ts
```diff
@@ -424,6 +504,13 @@ class ThundraWrapper {
         const endTime = Math.min(configEndTime, maxEndTime);
 
         return setTimeout(() => {
+            if (this.debuggerProxy) {
+                // Debugger proxy exists, let it know about the timeout
+                if (!this.debuggerProxy.killed) {
```
serkan-ozal: It might be safer to put process kill in a try/catch and assign `null` to `debuggerProxy` in finally like did at the end of `startDebuggerProxyIfAvailable`

src/ThundraWrapper.ts
```diff
@@ -218,22 +253,60 @@ class ThundraWrapper {
...
+                        }
+
+                        if (typeof mes === 'string') {
+                            const match = mes.match(BROKER_WS_HTTP_ERROR_PATTERN);
+                            if (match) {
+                                const errCode = Number(match[1]);
+                                const errMessage = BROKER_WS_HTTP_ERR_CODE_TO_MSG[errCode];
+
+                                if (errMessage) {
+                                    ThundraLogger.getInstance().error('Thundra Debugger: ' + errMessage);
```
serkan-ozal: I think, even if there is no match for message, as we accept is as error, it should be logged. ıf it matches pattern, we log with pre-defined messages, otherwise, log with raw message

src/index.ts
```diff
@@ -42,43 +42,51 @@ let tracer: ThundraTracer;
...
-        InvocationTraceSupport.tracer = tracer;
+    if (!(config.apiKey)) {
+        console.warn(`Thundra API Key is not given, instrumentation is disabled.`);
     }
 
-    if (!(Utils.getConfiguration(envVariableKeys.THUNDRA_DISABLE_METRIC, 'true') === 'true') && config.metricConfig.enabled) {
-        const metricPlugin = MetricPlugin(config.metricConfig);
-        config.plugins.push(metricPlugin);
-    }
+    const instrumentationDisabled: boolean = !(config.apiKey);
```
serkan-ozal: I would prefer `monitoringDisabled` term as it also disables metric and logging

src/index.ts
```diff
@@ -42,43 +42,51 @@ let tracer: ThundraTracer;
...
-        const tracePlugin = TracePlugin(config.traceConfig);
-        config.plugins.push(tracePlugin);
-
-        tracer = tracePlugin.tracer;
-        config.metricConfig.tracer = tracer;
-        config.logConfig.tracer = tracer;
-        config.xrayConfig.tracer = tracer;
-        InvocationTraceSupport.tracer = tracer;
+    if (!(config.apiKey)) {
+        console.warn(`Thundra API Key is not given, instrumentation is disabled.`);
```
serkan-ozal: I would use `Thundra API Key is not given, monitoring is disabled.` as error message as metric and logging is also disabled

PR #142, opened 2020-03-06
use data.Entries Ids prop for generating incoming trace link

fatihaydilek:


src/plugins/integrations/AWSIntegration.ts
```diff
@@ -335,9 +335,9 @@ class AWSIntegration implements Integration {
                     span.setTag(AwsAthenaTags.RESPONSE_NAMED_QUERY_IDS, [response.data.NamedQueryId]);
                 }
             } else if (serviceName === 'events') {
-                const eventId = get(response, 'id', false);
-                if (eventId) {
-                    traceLinks = [eventId];
+                const eventIds = get(response, 'data.Entries', []).map((e: any) => e.EventId);
+                if (eventIds) {
```
hamitb: The condition inside the if check always evaluates to true so we can get rid of the if check.

PR #154, opened 2020-04-10
Feature/zeit integration

hamitb:


src/Constants.ts
```diff
@@ -232,10 +232,16 @@ export const ClassNames = {
     ELASTICSEARCH: 'ELASTICSEARCH',
     MYSQL: 'MYSQL',
     EVENTBRIDGE: 'AWS-EventBridge',
+    ZEIT: 'ZEIT',
 };
 
 export const AWS_SERVICE_REQUEST = 'AWSServiceRequest';
 
+export const ZeitTags = {
+    DEPLOYMENT_URL_HEADER: 'x-now-deployment-url',
```
serkan-ozal, line 241: Can we have `ZeitConstants` and move `DEPLOYMENT_URL_HEADER` to there as `DEPLOYMENT_URL_HEADER` is not a tag but a constant

src/plugins/utils/LambdaEventUtils.ts
```diff
@@ -49,6 +49,15 @@ class LambdaEventUtils {
         } else if (originalEvent['detail-type'] && originalEvent.detail &&  originalEvent.version
             && Array.isArray(originalEvent.resources)) {
             return LambdaEventType.EventBridge;
+        } else if (originalEvent.Action && originalEvent.body) {
+            try {
+                const { headers } = JSON.parse(originalEvent.body);
+                if (headers[ZeitTags.DEPLOYMENT_URL_HEADER]) {
```
serkan-ozal, line 55: It might be better to check whether or not `headers` is exist in the body. 
```js
if (headers && headers[ZeitTags.DEPLOYMENT_URL_HEADER]) {
```
It will prevent unnecessary exceptions even though it is catched below. Because throwing exception is not free operation as it requires to collect stacktrace.

src/Constants.ts
hamitb: 👍

src/plugins/utils/LambdaEventUtils.ts
hamitb: 👍

PR #159, opened 2020-04-21
add default process.env.LAMBDA_TASK_ROOT as resolve path to require l…

fatihaydilek:
…ogic

### DIFF
--- a/src/plugins/utils/Utils.ts
+++ b/src/plugins/utils/Utils.ts
@@ -248,11 +248,7 @@ class Utils {
 
     static tryRequire(name: string, paths?: string[]): any {
         try {
-            if (paths === undefined) {
-                paths = [process.env.LAMBDA_TASK_ROOT];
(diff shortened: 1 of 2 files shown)


src/plugins/utils/Utils.ts
```diff
@@ -248,26 +248,25 @@ class Utils {
...
     static tryRequire(name: string, paths?: string[]): any {
         try {
-            let resolvedPath;
-            if (paths !== undefined) {
-                resolvedPath = customReq.resolve(name, { paths });
+            if (paths !== undefined && paths.indexOf(process.env.LAMBDA_TASK_ROOT) === -1) {
+                paths.push(process.env.LAMBDA_TASK_ROOT);
             } else {
-                resolvedPath = customReq.resolve(name);
+                paths = [process.env.LAMBDA_TASK_ROOT];
```
serkan-ozal, line 252: If `/var/task` is already in the paths with others, this else statement overrides all of them with only `/var/task`

PR #160, opened 2020-04-27
Update eventbrige span tags

fatihaydilek:


src/config/ConfigNames.ts
```diff
@@ -142,6 +142,9 @@ class ConfigNames {
     public static readonly THUNDRA_TRACE_INTEGRATIONS_MONGODB_COMMAND_MASK: string =
         'thundra.agent.trace.integrations.mongodb.command.mask';
 
+    public static readonly THUNDRA_TRACE_INTEGRATIONS_EVENTBRIDGE_DETAIL_MASK: string =
+        'thundra.agent.lambda.trace.integrations.aws.eventbridge.detail.mask';
```
serkan-ozal, line 146: This must be `thundra.agent.trace.integrations.aws.eventbridge.detail.mask`. No need to `lambda` as it is not specific to Lambda environment

PR #161, opened 2020-05-06
Instrument with shimmer and hook

fatihaydilek:


src/plugins/utils/Utils.ts
```diff
@@ -257,32 +261,96 @@ class Utils {
...
+                    Utils.doInstrument(requiredLib, libs, basedir, moduleName, version, wrapper, config);
+                } else {
+                    Utils.doInstrument(requiredLib, libs, null, moduleName, null, wrapper, config);
+                }
+            }
+            const hook = Hook(moduleName, {internals: true}, (lib: any, name: string, basedir: string) => {
+                if (name === moduleName) {
+                    Utils.doInstrument(lib, libs, basedir, moduleName, version, wrapper, config);
+                }
+                hooks.push(hook);
```
serkan-ozal, line 337: I think, this should be outside of Hook block as in all cases (even though instrumentation is not applied during module load) hooks need to be unhook. I mean
```js
            const hook = Hook(moduleName, {internals: true}, (lib: any, name: string, basedir: string) => {
                if (name === moduleName) {
                    Utils.doInstrument(lib, libs, basedir, moduleName, version, wrapper, config);
                }
                return lib;
            });
            hooks.push(hook);
```

PR #166, opened 2020-05-31
Got rid of `ThundraLogger.getInstance()` and accessed over static methods directly …

serkan-ozal:
… as no need to have a single instance of it

### DIFF
--- a/src/ThundraLogger.ts
+++ b/src/ThundraLogger.ts
@@ -2,39 +2,28 @@ import ConfigProvider from './config/ConfigProvider';
 import ConfigNames from './config/ConfigNames';
 
 class ThundraLogger {
-    static instance: ThundraLogger;
-    debugEnabled: boolean;
 
-    constructor() {
-        this.debugEnabled = ConfigProvider.get<boolean>(ConfigNames.THUNDRA_DEBUG_ENABLE);
-        ThundraLogger.instance = this;
+    private constructor() {
     }
 
-    static getInstance(): ThundraLogger {
-        return ThundraLogger.instance ? ThundraLogger.instance : new ThundraLogger();
+    static isDebugEnabled(): boolean {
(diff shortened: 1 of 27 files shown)


src/ThundraLogger.ts
```diff
@@ -2,39 +2,35 @@ import ConfigProvider from './config/ConfigProvider';
...
-    static instance: ThundraLogger;
-    debugEnabled: boolean;
 
-    constructor() {
-        this.debugEnabled = ConfigProvider.get<boolean>(ConfigNames.THUNDRA_DEBUG_ENABLE);
-        ThundraLogger.instance = this;
-    }
+    static debugEnabled: boolean;
 
-    static getInstance(): ThundraLogger {
```
hamitb, line 13: I didn't see where this constructor is called, so if this is the case, I think `static debugEnabled: boolean` field is always equal to the default value `undefined`.

src/ThundraLogger.ts
```diff
@@ -2,39 +2,35 @@ import ConfigProvider from './config/ConfigProvider';
...
-        return ThundraLogger.instance ? ThundraLogger.instance : new ThundraLogger();
+    private constructor() {
```
hamitb, line 8: I didn't see where this constructor is called, so if this is the case, I think `static debugEnabled: boolean` field is always equal to the default value `undefined`.

serkan-ozal: Yes, it was not called. Also since static initializer of `ThundraLogger` is called before `ConfigProvider` is initialized, I updated debug enable check by looking at config on `isDebugEnabled` method call. Otherwise, `ThundraLogger` was not getting the right config value and I also didn't want to tweak with the module initialization order

serkan-ozal: Also keep private constructor to prevent `ThundraLogger` from being instantiated

hamitb: 👍

PR #168, opened 2020-06-02
Refactor/aws integration

hamitb:


serkan-ozal:
@hamitb I think, it might be better to have another phase like `onResponse` so it will handle response (for ex. add some tags to span according to response from AWS service). And seems that currently only `AWSAthenaIntegration` will use it

src/plugins/integrations/AWSIntegration.ts
```diff
@@ -819,4 +257,756 @@ class AWSIntegration implements Integration {
...
+            }
+        }
+
+        return activeSpan;
+    }
+
+    public static createTraceLinks(span: ThundraSpan, request: any, config: any): any[] {
+        const response = request.response;
+
+        if (has(response, 'data.QueryExecutionIds')) {
```
serkan-ozal, line 330: I think, we should introduce another phase `onResponse` and the tags should be handled there

PR #169, opened 2020-06-05
[THN-2347] separate lambda related application info logic from index

bunyaminsg:


src/index.ts
```diff
@@ -91,36 +93,32 @@ module.exports = (options?: any) => {
...
-
     return (originalFunc: any) => {
         // Check if already wrapped
         if (get(originalFunc, 'thundraWrapped', false)) {
             return originalFunc;
         }
 
+        ApplicationManager.setApplicationInfoProvider(new LambdaApplicationInfoProvider());
+        const applicationInfoProvider = ApplicationManager.getApplicationInfoProvider();
+        const applicationInfo = applicationInfoProvider.getApplicationInfo();
```
hamitb, line 104: We can directly use `const applicationInfo = ApplicationManager.getApplicationInfo()` here, instead of accessing it through the `applicationInfoProvider`

hamitb, line 104: We can directly use `const applicationInfo = ApplicationManager.getApplicationInfo()` here, instead of accessing it through the `applicationInfoProvider`

serkan-ozal: make sense :+1:

PR #174, opened 2020-06-15
Add netlify trigger

hamitb:


src/plugins/utils/LambdaEventUtils.ts
```diff
@@ -360,6 +363,32 @@ class LambdaEventUtils {
         return className;
     }
 
+    static injectTriggerTagsForNetlify(span: ThundraSpan, pluginContext: any, originalContext: any): string {
+        const className = ClassNames.NETLIFY;
+        const domainName = DomainNames.API;
+        let operationName = 'netlify-site';
+
+        const siteName = process.env[NetlifyConstants.NETLIFY_SITE_NAME];
```
serkan-ozal, line 371: It is better to get env var over `Utils.getEnvVar` instead of dealing with `process.env` directly

src/plugins/utils/LambdaEventUtils.ts
```diff
@@ -360,6 +363,32 @@ class LambdaEventUtils {
...
+
+        if (siteName) {
+            operationName = siteName;
+        }
+
+        const originalHandler = process.env[EnvVariableKeys._HANDLER];
```
serkan-ozal, line 377: It is better to get env var over `Utils.getEnvVar` instead of dealing with `process.env` directly

src/plugins/utils/LambdaEventUtils.ts
```diff
@@ -360,6 +363,32 @@ class LambdaEventUtils {
...
+        const functionName = operationName + '/' + originalHandler.substring(0, originalHandler.indexOf('.'));
+
+        span._setOperationName(functionName);
+        originalContext.functionName = functionName;
```
serkan-ozal, line 381: Might there be a side effect as we are mutating original context property? Might there be another way?

src/plugins/utils/LambdaEventUtils.ts
```diff
@@ -360,6 +363,32 @@ class LambdaEventUtils {
...
+        let operationName = 'netlify-site';
+
+        const siteName = process.env[NetlifyConstants.NETLIFY_SITE_NAME];
+
+        if (siteName) {
+            operationName = siteName;
+        }
+
+        const originalHandler = process.env[EnvVariableKeys._HANDLER];
+        const functionName = operationName + '/' + originalHandler.substring(0, originalHandler.indexOf('.'));
```
serkan-ozal, line 378: I think, it is better to be at safe side here by checking whether original handler is available as env var and it has `.` character

src/plugins/utils/LambdaEventUtils.ts
serkan-ozal: Is it just needed to be used while getting application id over `LambdaUtils.getApplicationId`?

PR #178, opened 2020-06-23
Add data logs for debugger

hamitb:


src/debugBridge.ts
```diff
@@ -33,6 +33,7 @@ const brokerSocket = new WebSocket(
 
 // Setup debugger socket
 debuggerSocket.on('data', (data: Buffer) => {
+    log('from function: ', data.toString('utf8').substring(0, 50));
```
serkan-ozal, line 36: it might be better to put these log statement into if block:
```nodejs
if (LOGS_ENABLED === 'true') {
     log('from function: ', data.toString('utf8').substring(0, 50));
}
```
So there won't be extra overhead for `data.toString('utf8').substring(0, 50)` when logging is disabled

src/debugBridge.ts
```diff
@@ -64,7 +65,7 @@ brokerSocket.on('message', (data: Buffer) => {
             log('debuggerSocket: connection established with main lambda process');
         });
     }
-
+    log('from broker: ', data.toString('utf8').substring(0, 50));
```
serkan-ozal, line 68: it might be better to put these log statements into if block:
```js
if (LOGS_ENABLED === 'true') {
     log('from broker: ', data.toString('utf8').substring(0, 50));
}
```
So there won't be extra overhead for `data.toString('utf8').substring(0, 50)` when logging is disabled

PR #180, opened 2020-06-25
Integrate node agent with AWS SES

bunyaminsg:


src/Constants.ts
```diff
@@ -498,6 +511,7 @@ export const INTEGRATIONS: any = {
     aws: AWSIntegration,
     es: ESIntegration,
     mongodb: MongoDBIntegration,
+    ses: AWSSESIntegration,
```
serkan-ozal, line 514: This should not be needed. The main `AWSIntegration` is already there

src/Constants.ts
```diff
@@ -5,7 +5,7 @@ import MySQLIntegration from './plugins/integrations/MySQLIntegration';
 import RedisIntegration from './plugins/integrations/RedisIntegration';
 import MongoDBIntegration from './plugins/integrations/MongoDBIntegration';
 import IORedisIntegration from './plugins/integrations/IORedisIntegration';
-import { AWSIntegration } from './plugins/integrations/AWSIntegration';
+import {AWSIntegration, AWSSESIntegration} from './plugins/integrations/AWSIntegration';
```
serkan-ozal, line 8: `AWSSESIntegration` import should not be needed as no need to put `AWSSESIntegration` into `INTEGRATIONS` because the main `AWSIntegration` is already there

src/config/ConfigMetadata.ts
```diff
@@ -202,6 +202,18 @@ export const ConfigMetadata: {[key: string]: { type: string, defaultValue?: any
         type: 'boolean',
         defaultValue: false,
     },
+    [ConfigNames.THUNDRA_TRACE_INTEGRATIONS_AWS_SES_BODY_MASK]: {
```
serkan-ozal, line 205: I don't think we need to have two distinct configurations for masking body and subject separately. For simpler configuration we can have a single config `THUNDRA_TRACE_INTEGRATIONS_AWS_SES_MAIL_MASK`

serkan-ozal:
@bunyaminsg Can you add integration tests as for other AWS services

PR #183, opened 2020-07-20
Fix/pg sequielize

hamitb:


src/plugins/integrations/PostgreIntegration.ts
```diff
@@ -131,7 +110,40 @@ class PostgreIntegration implements Integration {
...
+                            span.closeWithCallback(me, originalCallback, [err, res]);
+                        };
+                        newArgs[callbackIndex] = wrappedCallback;
+                    }
+
+                    const result = query.apply(this, newArgs);
+
+                    if (result && typeof result.then === 'function') {
+                        result.then(function (value: any) {
+                            span.closeWithCallback(me , originalCallback, [null, value]);
```
serkan-ozal, line 138: As far as I see, when it returns promise, there will be no callback. So do we still need close with callback here? Can't we just close span here? Is it valid case to use callback even it expects promise?

PR #188, opened 2020-07-30
Step functions integration

hamitb:


src/wrappers/lambda/LambdaExecutor.ts
```diff
@@ -292,6 +300,32 @@ function injectTriggerTags(span: ThundraSpan, pluginContext: any, originalEvent:
...
+function injectStepFunctionInfo(request: any, response: any): any {
+    try {
+        const isStepFunction = ConfigProvider.get<boolean>(ConfigNames.THUNDRA_LAMBDA_AWS_STEPFUNCTIONS);
+
+        if (isStepFunction) {
+            const traceLink = Utils.generateId();
+            let step = 0;
+
+            if (THUNDRA_TRACE_LINK_KEY in request) {
+                step = request[THUNDRA_TRACE_LINK_KEY].step;
```
serkan-ozal, line 312: If `step` is not exist in the request, doesn't this make `step` undefined

src/Constants.ts
```diff
@@ -671,3 +681,5 @@ export const BROKER_WS_HTTP_ERR_CODE_TO_MSG: {[key: number]: string} = {
     401: `Authentication is failed, check your Thundra debugger authentication token.`,
     409: `Another session with the same session name exists, connection closed.`,
 };
+
+export const THUNDRA_TRACE_LINK_KEY = '_thundra';
```
serkan-ozal, line 685: I think, `THUNDRA_TRACE_KEY` is better name as in the future there might be additional data under `_thundra` key, not just trace link

src/wrappers/lambda/LambdaEventUtils.ts
```diff
@@ -21,6 +22,17 @@ class LambdaEventUtils {
...
     private constructor() {
     }
+    /**
+     * Checks if a propagated trace link exists in the incoming event
+     * @param originalEvent the original AWS Lambda invocation event
+     */
+    static extractTraceLinkFromEvent(originalEvent: any) {
+        try {
+            if (THUNDRA_TRACE_LINK_KEY in originalEvent) {
+                InvocationTraceSupport.addIncomingTraceLink(originalEvent[THUNDRA_TRACE_LINK_KEY].trace_link);
```
serkan-ozal, line 32: It might be better to check whether `trace_link` is really exist in the `_thundra` key

src/integrations/AWSIntegration.ts
```diff
@@ -616,6 +618,84 @@ export class AWSSNSIntegration {
...
+        try {
+            const originalInput = get(request, 'params.input');
+
+            if (originalInput) {
+                const parsedInput = JSON.parse(originalInput);
+                const traceLink = Utils.generateId();
+
+                parsedInput[THUNDRA_TRACE_LINK_KEY] = { trace_link: traceLink, step: 0 };
+
+                span.setTag(AwsStepFunctionsTags.EXECUTION_INPUT, originalInput);
```
serkan-ozal, line 688: I think, it is better to put this before `JSON.parse` above so even if the input is not in JSON format, we can still see the original input through tags

PR #202, opened 2020-08-04
Handle reporting errors

serkan-ozal:


src/wrappers/lambda/LambdaHandlerWrapper.ts
```diff
@@ -423,7 +423,14 @@ class LambdaHandlerWrapper {
                 this.reported = true;
                 this.destroyTimeoutHandler();
 
-                await this.executeAfterInvocationAndReport();
+                try {
+                    await this.executeAfterInvocationAndReport();
+                } catch (e) {
+                    if (ThundraLogger.isDebugEnabled()) {
```
hamitb, line 429: `ThundraLogger.debug()` is already making this check, so I think we don't need the `if` part.

serkan-ozal: right, `isDebugEnabled` check is used for non-cheap operations so I think it is better to remove it here

PR #230, opened 2020-08-17
Express integration

hamitb:


src/utils/Utils.ts
```diff
@@ -557,6 +578,18 @@ class Utils {
...
 
+    static getNormalizedPath(pathStr: string, depth: number): string {
+        try {
+            if (depth <= 0) {
+                return '';
+            }
+            const normalizedPath = '/' + pathStr.split('/').filter((c) => c !== '').slice(0, depth).join('/');
+            return normalizedPath;
+        } catch (error) {
+            return pathStr;
```
serkan-ozal, line 589: Can you add debug log here

src/wrappers/WebWrapperUtils.ts
```diff
@@ -0,0 +1,218 @@
...
+import TimeoutError from '../error/TimeoutError';
+import InvocationSupport from '../plugins/support/InvocationSupport';
+import InvocationTraceSupport from '../plugins/support/InvocationTraceSupport';
+import { HttpTags, DomainNames, ClassNames, SpanTags, TriggerHeaderTags } from '../Constants';
+import ThundraSpanContext from '../opentracing/SpanContext';
+
+const get = require('lodash.get');
+
+export default class WrapperUtils {
+    static beforeRequest(request: any, plugins: any[]) {
```
serkan-ozal, line 28: I think it is better to make `beforeRequest` async and await `beforeInvocation` calls of plugins as `beforeInvocation` phase might be async for some plugins such as metric. Even though currently metric plugin is not available for web wrapper, I think we should contact with plugins in async way.

src/wrappers/WebWrapperUtils.ts
```diff
@@ -0,0 +1,218 @@
...
+import InvocationData from '../plugins/data/invocation/InvocationData';
+import TimeoutError from '../error/TimeoutError';
+import InvocationSupport from '../plugins/support/InvocationSupport';
+import InvocationTraceSupport from '../plugins/support/InvocationTraceSupport';
+import { HttpTags, DomainNames, ClassNames, SpanTags, TriggerHeaderTags } from '../Constants';
+import ThundraSpanContext from '../opentracing/SpanContext';
+
+const get = require('lodash.get');
+
+export default class WrapperUtils {
```
serkan-ozal, line 27: It is better to rename to `WebWrapperUtils`

src/wrappers/WebWrapperUtils.ts
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    static async afterRequest(response: any, plugins: any[], reporter: Reporter) {
+        const context = ExecutionContextManager.get();
+
+        context.finishTimestamp = Date.now();
+        context.response = response;
+
+        for (const plugin of plugins) {
+            plugin.afterInvocation(context);
```
serkan-ozal, line 46: await here

src/wrappers/WebWrapperUtils.ts
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        try {
+            await reporter.sendReports(context.reports);
+        } catch (err) {
+            ThundraLogger.error(err);
```
serkan-ozal, line 52: Can you log error in `WebWrapperUtils` context by 
```js
ThundraLogger.error('<WebWrapperUtils>:', err);
```

serkan-ozal: Should we really wait for sending reports here?

src/wrappers/express/ExpressWrapper.ts
```diff
@@ -0,0 +1,39 @@
...
+    return (req: any, res: any, next: any) => ExecutionContextManager.runWithContext(
+        WrapperUtils.createExecContext,
+        () => {
+            try {
+                WrapperUtils.beforeRequest(req, plugins);
+                res.once('finish', () => {
+                    WrapperUtils.afterRequest(res, plugins, reporter);
+                });
+            } catch (err) {
+                console.error(err);
```
serkan-ozal, line 33: This should be replaced by `ThundraLogger`

src/wrappers/express/ExpressWrapper.ts
```diff
@@ -0,0 +1,43 @@
...
+        async function () {
+            const context: ExecutionContext = this;
+            try {
+                await WrapperUtils.beforeRequest(req, plugins);
+                res.once('finish', () => {
+                    ExecutionContextManager.set(context);
+                    WrapperUtils.afterRequest(res, plugins, reporter);
+                });
+            } catch (err) {
+                ThundraLogger.error('<ExpressWrapper> Error occured in in ExpressWrapper: ', err);
```
serkan-ozal, line 37: - There are multiple `in` in the log message
- No need to extra space after `:` as space is already added automatically after each argument

src/wrappers/express/ExpressWrapper.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (!opts.disableAsyncContextManager) {
+        WrapperUtils.initAsyncContextManager();
+    }
+
+    return (req: any, res: any, next: any) => ExecutionContextManager.runWithContext(
+        WrapperUtils.createExecContext,
+        async function () {
+            const context: ExecutionContext = this;
+            try {
+                await WrapperUtils.beforeRequest(req, plugins);
```
serkan-ozal, line 31: I think it might be better to to set a flag after `await WrapperUtils.beforeRequest(req, plugins);` so it indicates that before request completed successfully. And then `WrapperUtils.afterRequest(res, plugins, reporter);` is executed only if this flag is set. So if somehow before request fails, after request is not executed

serkan-ozal: Ok, never mind this comment

serkan-ozal (approved):

serkan-ozal (approved):

PR #232, opened 2020-08-26
Fix/vercel root span operation name

fatihaydilek:


src/wrappers/lambda/LambdaEventUtils.ts
```diff
@@ -473,11 +473,11 @@ class LambdaEventUtils {
                     applicationName,
                     applicationId: LambdaPlatformUtils.getApplicationId(originalContext, {functionName: applicationName}),
                 });
+                span.setOperationName(path || '');
```
serkan-ozal, line 476: It might be better to keep existing operation name if there is no path.
```js
if (path) {
    span.setOperationName(path);
}
```

serkan-ozal (approved):

PR #244, opened 2020-10-09
Add application resource name

hamitb:


src/application/ApplicationInfo.ts
```diff
@@ -12,5 +11,5 @@ export interface ApplicationInfo {
     applicationVersion: string;
     applicationStage: string;
     applicationTags: any;
-
+    applicationResourceName: string;
```
serkan-ozal, line 14: As application resource name might be different for each execution (for ex, same web app handle requests concurrently for different endpoints),  `applicationResourceName` should not be here

src/wrappers/express/ExpressExecutor.ts
```diff
@@ -65,15 +66,18 @@ export function finishTrace(pluginContext: PluginContext, execContext: Execution
...
 
-    const triggerOperationName = request.hostname + resourceName;
+    const triggerOperationName = resourceName;
     context.triggerOperationName = triggerOperationName;
 
     // Change root span name and response header
     rootSpan.operationName = resourceName;
     response.set(TriggerHeaderTags.RESOURCE_NAME, triggerOperationName);
     InvocationSupport.setAgentTag(SpanTags.TRIGGER_OPERATION_NAMES, [triggerOperationName]);
+    ApplicationManager.getApplicationInfoProvider().update({
```
serkan-ozal, line 78: I think, `applicationResourceName` should be stored into `ExecutionContext` and invocation data should be feed from there

serkan-ozal (approved):

PR #247, opened 2020-10-16
fix functionName and functionVersion construction for AWS Greengrass

aoozdemir:


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@thundra/core",
-  "version": "2.12.3",
+  "version": "2.12.4",
```
aoozdemir, line 3: Bunu otomatik yapti, acaba bu degisikligi de pushlamama gerek var mi?

serkan-ozal: @aoozdemir it is not needed. It is automatically increased while releasing

aoozdemir: @serkan-ozal Ah, I see what happened. `2.14.4` is already released but the lock file hasn't been updated. So, I guess it's better I keep this change, right?

serkan-ozal: Yes, seems that synchronization between package.json and package-lock.json was somehow broken by a manual change. So it is OK to change it manually `2.14.4` here

src/utils/Utils.ts
```diff
@@ -592,6 +593,25 @@ class Utils {
         }
     }
 
+    static getNormalizedFunctionName(request: any, lambdaArn?: string) {
```
serkan-ozal, line 596: I think it might be better to have `LambdaUtils` and move `getNormalizedFunctionName` to there. Otherwise, with the new platform supports, `Utils` might be messed up

aoozdemir: @serkan-ozal makes sense. I'll do the appropriate changes.

serkan-ozal (approved):

PR #249, opened 2020-10-30
write and include the initial index.d.ts file

rwxdash:


src/index.d.ts
```diff
@@ -0,0 +1,143 @@
...
+   * Gets the tracer
+   * @return {Tracer} the tracer
+   */
+  export function tracer(): ThundraTracer;
+
+  /**
+   * Loads and returns the user AWS Lambda handler
+   * @return the loaded user AWS Lambda handler
+   */
+  export function loadUserHandler(): any;
```
aoozdemir, line 57: I'm not sure about this one. `loadUserHandler` calls the `loadHandler`. However, `loadHandler` doesn't have any return value in the code, which confused me a little. Giving `any` as a return value something I didn't want to do. Open to suggestions.

src/index.d.ts
```diff
@@ -0,0 +1,143 @@
...
+   * implementation.
+   *
+   * The default object acts as a no-op implementation.
+   *
+   * Note to implementators: derived classes can choose to directly implement the
+   * methods in the "OpenTracing API methods" section, or optionally the subset of
+   * underscore-prefixed methods to pick up the argument checking and handling
+   * automatically from the base class.
+   */
+  export class Tracer { }
```
aoozdemir, line 140: These classes, `ThundraTracer` and `Tracer`, need some dependencies before I include the definitions, but given that including those dependency files could increase our size and complexity, I've decided to leave it empty. 

See `tracer()` function.

PR #254, opened 2020-11-26
Improvement/express/handle init

fatihaydilek:


src/wrappers/WebWrapperUtils.ts
```diff
@@ -44,6 +44,10 @@ export default class WebWrapperUtils {
     static async afterRequest(request: any, response: any, plugins: any[], reporter: Reporter) {
         const context = ExecutionContextManager.get();
 
+        if (!context.error && response.statusCode >= 500) {
+            context.error = Utils.buildError(new Error(response.statusMessage));
```
serkan-ozal, line 48: I think, here setting `HTTPError` might be better to be consistent with Lambda part.
```js
context.error = new HttpError(`Returned with error status code ${response.statusCode}.`);
```

serkan-ozal (approved):

PR #267, opened 2021-03-03
Fix/run all tests

fatihaydilek:


.github/workflows/build.yml
```diff
@@ -0,0 +1,28 @@
+name: Thundra CI Check
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches:
+      - "fix/run-all-tests"
```
aoozdemir, line 8: I think we can remove `"fix/run-all-tests"` from the PR triggers

PR #279, opened 2021-07-28
Ignore reporting child spans after AWS operation span

serkan-ozal:


src/opentracing/Tracer.ts
```diff
@@ -155,13 +155,24 @@ class ThundraTracer extends Tracer {
...
             throw new Error('Invalid spanId : ' + parentContext.spanId);
         }
 
+        const stopReporting: boolean = fields.stopReporting;
+        // Skip reporting if
+        // - it is explicitly specified
+        // - or recording of its parent is skipped
+        // - or recording is stopped from its parent
+        const skipReporting: boolean =
+            (fields && fields.skipReporting)
```
bcaglayan, line 171: null check does not necessary for 'fields' in this line, it can be removed or it can be placed before first usage

bcaglayan (approved):

src/opentracing/Tracer.ts
serkan-ozal: removed that code part as it was not necessary for our new solution

fatihaydilek (approved):

PR #284, opened 2021-08-19
Amqlib integration

gokhan721:


src/Constants.ts
```diff
@@ -430,6 +432,19 @@ export const ESTags = {
...
+    [key: string]: string;
+}
+
+export const AMQPTags: AMQPLIBType = {
+    HOST: 'amqp.host',
+    PORT: 'amqp.port',
+    QUEUE: 'amqp.queue',
+    EXCHANGE: 'amqp.exchange',
+    ROUTING_KEY: 'amqp.routingKey',
+    CONSUMER_TAG: 'amqp.consumerTag',
```
serkan-ozal, line 445: No need to consumer tag for now

src/Constants.ts
```diff
@@ -430,6 +432,19 @@ export const ESTags = {
...
+interface AMQPLIBType {
+    [key: string]: string;
+}
+
+export const AMQPTags: AMQPLIBType = {
+    HOST: 'amqp.host',
+    PORT: 'amqp.port',
+    QUEUE: 'amqp.queue',
+    EXCHANGE: 'amqp.exchange',
+    ROUTING_KEY: 'amqp.routingKey',
```
serkan-ozal, line 444: `amqp.routing_key`

src/Constants.ts
```diff
@@ -430,6 +432,19 @@ export const ESTags = {
     ES_BODY: 'elasticsearch.body',
 };
 
+interface AMQPLIBType {
+    [key: string]: string;
+}
+
+export const AMQPTags: AMQPLIBType = {
```
serkan-ozal, line 439: We should also have a tag for the message itself: `amqp.message`

src/integrations/AmqplibIntegration.ts
```diff
@@ -0,0 +1,258 @@
...
+          } else {
+            return sendMessage.apply(this, args);
+          }
+        } finally {
+          span.finish();
+        }
+      };
+    }
+
+    function wrapDispatchMessage(dispatchMessage: Function) {
```
serkan-ozal, line 103: For this PR, we can skip tracing dispatch for now

src/integrations/AmqplibIntegration.ts
```diff
@@ -0,0 +1,258 @@
...
+        }
+      };
+    }
+
+    ThundraLogger.debug(
+      '<AMQPLIBIntegration> Wrapping Channel.sendMessage, \
+	 		Channel.sendImmediately, BaseChannel.dispatchMessage',
+    );
+    shimmer.wrap(lib.Channel.prototype, 'sendMessage', wrapSendMessage);
+    shimmer.wrap(lib.BaseChannel.prototype, 'dispatchMessage', wrapDispatchMessage);
```
serkan-ozal, line 163: No need to instrument dispatch for this PR

src/integrations/AmqplibIntegration.ts
```diff
@@ -0,0 +1,258 @@
...
+   */
+  doUnwrap(lib: any) {
+    ThundraLogger.debug('<AMQPLIBIntegration> Do unwrap');
+
+    ThundraLogger.debug(
+      '<AMQPLIBIntegration> Unwrapping Channel.sendMessage, Channel.sendImmediately, BaseChannel.dispatchMessage',
+    );
+
+    shimmer.unwrap(lib.Channel.prototype, 'sendMessage');
+    shimmer.unwrap(lib.BaseChannel.prototype, 'dispatchMessage');
```
serkan-ozal, line 177: No need to instrument dispatch for this PR

src/integrations/AmqplibIntegration.ts
```diff
@@ -0,0 +1,258 @@
...
+          ThundraLogger.error(
+            '<AMQPLIBIntegration> Error occurred while tracing AMQP ${method} method:',
+            error,
+          );
+
+          if (span) {
+            ThundraLogger.debug(
+              `<AMQPLIBIntegration> Because of error, closing AMQP span with name ${span.getOperationName()}`,
+            );
+            span.finish();
```
serkan-ozal, line 89: no need finish here as there is already `span.finish()` in the finally block

src/integrations/AmqplibIntegration.ts
```diff
@@ -0,0 +1,258 @@
...
+        tracer.inject(span.spanContext, opentracing.FORMAT_TEXT_MAP, properties.headers);
+        try {
+          return sendMessage.apply(this, args);
+        } catch (error) {
+          ThundraLogger.error(
+            '<AMQPLIBIntegration> Error occurred while tracing AMQP ${method} method:',
+            error,
+          );
+
+          if (span) {
```
serkan-ozal, line 85: set error to span by `span.setErrorTag(error)`

src/integrations/AmqplibIntegration.ts
serkan-ozal: `span.finish()` must be replaced with `span.close()`

src/integrations/AmqplibIntegration.ts
```diff
@@ -0,0 +1,258 @@
...
+            ThundraLogger.debug(
+              `<AMQPLIBIntegration> Because of error, closing AMQP span with name ${span.getOperationName()}`,
+            );
+            span.finish();
+          }
+
+          if (error instanceof ThundraChaosError) {
+            throw error;
+          } else {
+            return sendMessage.apply(this, args);
```
serkan-ozal, line 95: We should not call original method here. Because when the code execution comes here, this means that `sendMessage.apply(this, args);` has failed above. So if we call it again here in case of error, this means that we are trying to send message twice. So, lets just throw error here

serkan-ozal: In overall, I mean:
```js
    function wrapSendMessage(sendMessage: Function) {
      /**
       * Wrap the sendMessage
       * @param args sendMessage function parameters in order fields, properties and content.
       */
      return function sendMessageWithTrace(...args: any) {
        ThundraLogger.debug(
          `<AMQPLIBIntegration> Tracing sendMessage args: ${args}`,
        );
        let span;
        try {
          const {tracer} = ExecutionContextManager.get();

          if (!tracer) {
            ThundraLogger.debug('<AMQPLIBIntegration> Skipped tracing command as no tracer is available');
            return sendMessage.apply(this, args);
          }

          const method = 'basic.publish';
          const [fields, properties, content] = args;
          const parentSpan = tracer.getActiveSpan();
          span = tracer._startSpan(fields.routingKey, {
            childOf: parentSpan,
            domainName: DomainNames.MESSAGING,
            className: ClassNames.AMQP,
            disableActiveStart: true,
          });
          integration.handleTags(this, config, span, method, fields);
          span._initialized();
          tracer.inject(span.spanContext, opentracing.FORMAT_TEXT_MAP, properties.headers);
        } catch (error) {
          ThundraLogger.error(
              '<AMQPLIBIntegration> Error occurred while tracing AMQP ${method} method:',
              error,
          );
          if (span) {
            ThundraLogger.debug(
                `<AMQPLIBIntegration> Because of error, closing AMQP span with name ${span.getOperationName()}`,
            );
            span.setErrorTag(error);
            span.close();
            span = null;
          }
          if (error instanceof ThundraChaosError) {
            throw error;
          }
        }
        try {
          return sendMessage.apply(this, args);
        } catch (error) {
          if (span) {
            span.setErrorTag(error);
          }
        } finally {
          if (span) {
            span.close();
          }
        }
      };
    }
```

src/Constants.ts
gokhan721: done.

src/integrations/AmqplibIntegration.ts
gokhan721: done.

src/Constants.ts
gokhan721: done.

PR #285, opened 2021-09-08
Feature/jest integration

bcaglayan:
Feature/jest integration


src/plugins/config/ThundraConfig.ts
```diff
@@ -45,6 +47,12 @@ class ThundraConfig {
         this.warmupAware = ConfigProvider.get<boolean>(
             ConfigNames.THUNDRA_LAMBDA_WARMUP_WARMUPAWARE,
             options.warmupAware);
+        this.testProjectId = ConfigProvider.get<string>(
+            ConfigNames.THUNDRA_AGENT_TEST_PROJECT_ID,
+            options.testProjectId);
+        this.testRundId = ConfigProvider.get<string>(
```
serkan-ozal, line 53: typo, it should be `testRunId`

src/utils/TestRunnerUtils.ts
```diff
@@ -0,0 +1,34 @@
...
+        return Utils.generareIdFrom(testRunIdSeed);
+    }
+
+    static getDefaultTestRunId(
+        environment: string,
+        repoURL: string,
+        commitHash: string) {
+
+            /** todo: generate more unique id from parente process ? */
+            const runId = process.ppid + '_';
```
serkan-ozal, line 23: Can we use globally generated UUID here and return it? I mean generate a uuid, store it in a global/static variable and use it here. So we will have same UUID during the same test process. 

However, if Jest loads Thundra module multiple times for each test-case environment (so in this case, global/static variable is re-initialized multiple times for each test), we may use process env vars as global storage here. I mean
```js
    static getDefaultTestRunId(
        environment: string,
        repoURL: string,
        commitHash: string) {
            let defaultTestRunId = process.env["THUNDRA_AGENT_DEFAULT_TEST_RUN_ID"]
            if (!defaultTestRunId) {
                defaultTestRunId = uuid.v4();
                process.env["THUNDRA_AGENT_DEFAULT_TEST_RUN_ID"] = defaultTestRunId;
            }
            
            return TestRunnerUtils.getTestRunId(
                environment,
                repoURL,
                commitHash,
                defaultTestRunId,
            );
    }
```

src/utils/TestRunnerUtils.ts
```diff
@@ -0,0 +1,34 @@
+import Utils from './Utils';
+
+class TestRunnerUtils {
```
serkan-ozal, line 3: JS doc here

src/utils/Utils.ts
```diff
@@ -45,6 +49,10 @@ class Utils {
         return uuidv4();
     }
 
+    static generareIdFrom(value: string): string {
```
serkan-ozal, line 52: JS doc here

src/wrappers/WrapperContext.ts
```diff
@@ -0,0 +1,19 @@
+
+import Reporter from '../Reporter';
+import PluginContext from '../plugins/PluginContext';
+
+export default class WrapperContext {
```
serkan-ozal, line 5: JS doc here

src/wrappers/foresight/TestRunnerSupport.ts
```diff
@@ -0,0 +1,179 @@
...
+            .withSuccessfulCount(testRunScope.context.successfulCount)
+            .withFailedCount(testRunScope.context.failedCount)
+            .withIgnoredCount(testRunScope.context.ignoredCount)
+            .withAbortedCount(testRunScope.context.abortedCount)
+            .withHostName(hostName)
+            .withEnvironmentInfo(EnvironmentSupport.getEnvironmentInfo())
+        .build();
+
+    await wrapperContext.reporter.sendReports([Utils.generateReport(testRunStatus, apiKey)]);
+    startTestRunStatusEvent();
```
serkan-ozal, line 105: If reporting is failed somehow, sending testrun-status events is not scheduled here. So it might be better to `startTestRunStatusEvent()` into finally block:
```js
try {
    await wrapperContext.reporter.sendReports([Utils.generateReport(testRunStatus, apiKey)]);
} finally {
    startTestRunStatusEvent();
}
```
- Schedule `startTestRunStatusEvent` with

src/wrappers/foresight/TestRunnerSupport.ts
```diff
@@ -0,0 +1,179 @@
...
+export const startTestRunStatusEvent = () => {
+
+    if (testStatusReportFreq) {
+        finishTestRunStatusEvent();
+        testRunStatusWork = setTimeout(sendTestRunStatus, testStatusReportFreq);
+    }
+};
+
+export const finishTestRunStatusEvent = () => {
+    clearInterval(testRunStatusWork);
```
serkan-ozal, line 117: we might skip here if testRunStatusWork is available.
``` js
export const finishTestRunStatusEvent = () => {
    if (testRunStatusWork) {
        clearInterval(testRunStatusWork);
        testRunStatusWork = null;
    }
}
```

src/wrappers/foresight/lib/handlers/AfterEach.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    if (!span) {
+
+        ThundraLogger.debug('<AfterEach><finish process> Span can not be empty.');
+        return;
+    }
+
+    span.tags[TestRunnerTags.TEST_NAME] = event.testName;
+
+    HandlerUtils.finishSpanForTest(span, TestRunnerTags.TEST_AFTER_EACH_DURATION);
```
serkan-ozal, line 55: should we clear the span (`span = null;`) here after we are done with it.

src/wrappers/foresight/lib/handlers/AfterAll.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Finish for handling afterAll event
+ * @param event event
+ */
+export async function finish(event: TestSuiteEvent) {
+
+    ThundraLogger.debug('<AfterAll> Handling afterAll finish event.');
+
+    HandlerUtils.finishSpanForTest(span, TestRunnerTags.TEST_AFTER_ALL_DURATION);
```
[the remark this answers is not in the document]
serkan-ozal: should we check whether the span is not `null` here

src/wrappers/foresight/lib/handlers/AfterEach.ts
serkan-ozal: should we check whether the span is not `null` here

src/wrappers/foresight/lib/handlers/BeforeAll.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Finish for handling beforeAll event
+ * @param event event
+ */
+export async function finish(event: TestSuiteEvent) {
+
+    ThundraLogger.debug('<beforeAll> Handling beforeAll finish event.');
+
+    HandlerUtils.finishSpanForTest(span, TestRunnerTags.TEST_BEFORE_ALL_DURATION);
```
serkan-ozal, line 47: should we clear the span (`span = null;`) here after we are done with it.

serkan-ozal: should we check whether the span is not `null` here

src/wrappers/foresight/lib/handlers/BeforeEach.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    if (!span) {
+
+        ThundraLogger.debug('<BeforeEach><finish process> Span can not be empty.');
+        return;
+    }
+
+    span.tags[TestRunnerTags.TEST_NAME] = event.testName;
+
+    HandlerUtils.finishSpanForTest(span, TestRunnerTags.TEST_BEFORE_EACH_DURATION);
```
serkan-ozal, line 56: should we clear the span (`span = null;`) here after we are done with it.

src/wrappers/foresight/lib/handlers/Setup.ts
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    try {
+
+        await sendData(testRunStart);
+        ThundraLogger.debug(`
+            <Setup> Test run start event sended for test suite: ${TestRunnerSupport.testSuiteName}
+            with test run id: ${testRunStart.id}
+        `);
+        TestRunnerSupport.startTestRunStatusEvent();
```
serkan-ozal, line 39: I think, we should start sending testrun status events even though testrun start event fails

src/wrappers/foresight/model/TestRunMonitoringData.ts
```diff
@@ -0,0 +1,9 @@
+export default class TestRunMonitoringData {
+    dataModelVersion: string = '1.0';
+    agentVersion: string = '2.12.51'; // todo: obtain agent version, impl a logic for it
```
serkan-ozal, line 3: You can get the agent version from `Constants.AGENT_VERSION`

src/utils/TestRunnerUtils.ts
bcaglayan: process env which are set inside a test-case environment are not visible each other. I mean 

`process.env["THUNDRA_AGENT_DEFAULT_TEST_RUN_ID"] `

this is not in global storage, so it is empty each time and it will be re-initialized multiple times. This usage causes to generate different testRundId for each test-suite.

src/bootstrap/index.ts
```diff
@@ -0,0 +1,3 @@
+const Thundra = require('@thundra/core');
+
+Thundra.foresightInit();
```
serkan-ozal, line 3: Since this is Foresight specific bootstrap, maybe we should put the `index.ts` file under `bootstrap/foresight/ `folder (and so of course move `bootstrap/jest` folder to `bootstrap/foresight/ jest` folder)

src/context/ExecutionContextManager.ts
```diff
@@ -21,6 +22,16 @@ export default class ExecutionContextManager {
      * @param provided the {@link ExecutionContext} provider to be set
      */
     static setProvider(provider: any) {
+
+        const contextCanChangeable = contextProvider.canChangeablebleContext();
+        if (!contextCanChangeable) {
+
+            ThundraLogger.debug('<ExecutionContextManager> Default context provider set as ForesightContextProvider.'
```
serkan-ozal, line 29: I think it might be better to write `contextProvider` name (maybe we should add name to context provider) here instead of using `ForesightContextProvider` as constant as in theory it might be another non-changable context

src/utils/HTTPUtils.ts
```diff
@@ -13,6 +14,37 @@ class HTTPUtils {
...
+        if (!path || !socketPath) {
+            return false;
+        }
+
+        if (host !== 'localhost'
+            || socketPath !== '/var/run/docker.sock') {
+            return false;
+        }
+
+        if (TESTCONTAINERS_HTTP_PATH_PATTERNS.PATTERN1.test(path)
```
serkan-ozal, line 37: I am not sure whether it is possible that we might ignore non-test container requests because of first fragment of the path. Doesn't ignoring `/var/run/docker.sock` is necessary?

src/wrappers/foresight/lib/jest/ModuleLoader.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    loadIntegrations(testRequire);
+    loadWrappers(testRequire);
+};
+
+const loadIntegrations = (testRequire: any) => {
+
+    for (const key in INTEGRATIONS) {
+        const integration = INTEGRATIONS[key];
+        if (integration && !integration.buildInIntegration) {
```
serkan-ozal, line 22: I think, `buildInIntegration` configuration and its check here should not be needed as such modules (http, https) are already ignored by ModuleUtils if they are retried to be instrumented

src/wrappers/foresight/TestRunnerSupport.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    const {
+        totalCount,
+        successfulCount,
+        failedCount,
+        ignoredCount,
+        abortedCount,
+    } = testRunScope.context;
+
+    return TestRunFinish.
```
serkan-ozal, line 177: clear (set `null`) `testRunScope`?

src/wrappers/foresight/context/ForesightContextProvider.ts
```diff
@@ -0,0 +1,114 @@
...
+export function canChangeablebleContext() {
+    return false;
+}
+
+export function runWithContext(createExecContext: Function, fn: Function) {
+
+    const execContext: ExecutionContext = createExecContext();
+
+    if (activeGlobalExecutionContext != null
+        && execContext.isContextCompatibleWith(ContextMode.AsyncMode)) {
```
serkan-ozal, line 78:
```js
execContext.isContextCompatibleWith(ContextMode.GlobalMode)
```
???

src/wrappers/foresight/sampler/MaxCountAwareSampler.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Sampler for check max count
+ */
+export default class MaxCountAwareSampler implements Sampler<null> {
+
+    private count: number;
+    private counter: number = 0;
+
+    constructor(count?: number) {
```
serkan-ozal, line 11: I think `?` should not be there as it is required

src/wrappers/foresight/TestRunnerSupport.ts
bcaglayan: clearTestRun method created and called after global teardown process

PR #303, opened 2021-11-02
Add config check for wrapping transform process

bcaglayan:
Add config check for wrapping transform process


src/wrappers/foresight/lib/jest/ModuleLoader.ts
```diff
@@ -9,6 +9,12 @@ import Trace from '../../../../plugins/Trace';
 const shimmer = require('shimmer');
 const has = require('lodash.has');
 
+export let TransformWrapped = false;
```
serkan-ozal, line 12: it might be better to export `isTransformWrapped` instead of exporting global `TransformWrapped` variable here

serkan-ozal (approved):

PR #323, opened 2021-12-12
Generic Function Wrapper

bcaglayan:
Generic Function Wrapper

### DIFF
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -114,6 +114,7 @@ export const DomainNames = {
     CACHE: 'Cache',
     SCHEDULE: 'Schedule',
     CDN: 'CDN',
+    NODE_GENERIC: 'Node-Generic',
 };
 
 export const ClassNames = {
@@ -143,6 +144,7 @@ export const ClassNames = {
     EXPRESS: 'Express',
     KOA: 'Koa',
     HAPI: 'Hapi',
+    NODE_HANDLER: 'Node-Handler',
     GOOGLE_BIGQUERY: 'Google-BigQuery',
     GOOGLE_PUBSUB: 'Google-PubSub',
     GOOGLE_PUBSUB_NODE_HANDLER: 'Google-PubSub-Node-Handler',
@@ -150,6 +152,10 @@ export const ClassNames = {
 };
 
 export const AWS_SERVICE_REQUEST = 'AWSServiceRequest';
+
+export const GenericWrapperTags = {
+    Function_Name: 'generic.wrapper.function.name',
+};
 
 export const AwsTags = {
     AWS_REGION: 'aws.region',
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -154,7 +154,7 @@ export const ClassNames = {
 export const AWS_SERVICE_REQUEST = 'AWSServiceRequest';
 
 export const GenericWrapperTags = {
-    Function_Name: 'generic.wrapper.function.name',
+    FUNCTION_NAME: 'generic.wrapper.function.name',
 };
 
 export const AwsTags = {
--- a/src/index.ts
+++ b/src/index.ts
@@ -166,7 +166,7 @@ if (global.__thundraMasterModule__) {
         listeners,
         init,
         Foresight,
-        GenericWrapper,
+        nodeWrapper: GenericWrapper.wrapper,
         updateConfig,
         createLogger,
         loadUserHandler,
--- a/src/index.ts
+++ b/src/index.ts
@@ -20,6 +20,7 @@ import LogManager from './plugins/LogManager';
 import InitManager from './init/InitManager';
 import ModuleUtils from './utils/ModuleUtils';
 import * as Foresight from './foresight';
+import * as GenericWrapper from './wrappers/generic/GenericWrapper';
 
 // Check if multiple instances of package exist
 if (!global.__thundraImports__) {
@@ -165,6 +166,7 @@ if (global.__thundraMasterModule__) {
         listeners,
         init,
         Foresight,
+        GenericWrapper,
         updateConfig,
         createLogger,
         loadUserHandler,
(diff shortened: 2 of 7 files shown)


src/wrappers/generic/GenericWrapper.ts
```diff
@@ -0,0 +1,113 @@
...
+        try {
+            ThundraLogger.debug('<GenericWrapper> Before handling request');
+
+            await beforeRequest(func);
+            beforeRequestCalled = true;
+
+            originalFunction = func(...args);
+            if (originalFunction.then) {
+                asyncFunction = true;
+                originalFunction.then(async (...data: any[]) => {
```
serkan-ozal, line 78: seems that `...data: any[]` is redundant here

src/wrappers/generic/GenericWrapper.ts
```diff
@@ -0,0 +1,113 @@
...
+        let asyncFunction: boolean = false;
+        let originalFunction;
+
+        try {
+            ThundraLogger.debug('<GenericWrapper> Before handling request');
+
+            await beforeRequest(func);
+            beforeRequestCalled = true;
+
+            originalFunction = func(...args);
```
serkan-ozal, line 75: `originalFunction` => `result`?

src/wrappers/generic/GenericWrapper.ts
```diff
@@ -0,0 +1,113 @@
...
+            if (originalFunction.then) {
```
serkan-ozal, line 76: `originalFunction.then` => `result && result.then && typeof result.then === 'function'`
Because
- return value might be undefined
- better to validate that `then` is a function

src/wrappers/generic/GenericExecutor.ts
```diff
@@ -0,0 +1,79 @@
...
+ * @param {ExecutionContext} execContext
+ */
+export function startTrace(pluginContext: PluginContext, execContext: ExecutionContext) {
+
+    const {
+        request,
+        tracer,
+    } = execContext;
+
+    const functionName = get(request, `functionName`);
```
serkan-ozal, line 47: Can function name be empty? for ex. anonymous functions?

src/wrappers/generic/GenericExecutor.ts
```diff
@@ -0,0 +1,79 @@
...
+    const contextInformation: any = execContext.getContextInformation();
+    const traceId = Utils.generateId();
+
+    const rootSpan = tracer._startSpan(functionName, {
+        propagated: false,
+        rootTraceId: Utils.generateId(),
```
serkan-ozal, line 53: No need to specify `rootTraceId` explicitly as default behaviour is already `Utils.generateId()`

src/wrappers/generic/GenericExecutor.ts
```diff
@@ -0,0 +1,79 @@
...
+        request,
+        tracer,
+    } = execContext;
+
+    const functionName = get(request, `functionName`);
+    const contextInformation: any = execContext.getContextInformation();
+    const traceId = Utils.generateId();
+
+    const rootSpan = tracer._startSpan(functionName, {
+        propagated: false,
```
serkan-ozal, line 52: No need to pass `propagated` flag as false explicitly here

src/Constants.ts
```diff
@@ -151,6 +153,10 @@ export const ClassNames = {
 
 export const AWS_SERVICE_REQUEST = 'AWSServiceRequest';
 
+export const GenericWrapperTags = {
+    Function_Name: 'generic.wrapper.function.name',
```
serkan-ozal, line 157: `Function_Name` => `FUNCTION_NAME`
Better to use upper-case letters to be consistent with others

src/wrappers/generic/GenericExecutor.ts
bcaglayan: Yes, it can.  What should our behavior be in this situation ? maybe, we can give default value like thundra-function ?

bcaglayan: On current behavior, tags value will be empy string. ( "" )

src/index.ts
```diff
@@ -165,6 +166,7 @@ if (global.__thundraMasterModule__) {
         listeners,
         init,
         Foresight,
+        GenericWrapper,
```
serkan-ozal, line 169: maybe it is better just export `wapper` method of `GenericWrapper` instead of exporting `GenericWrapper` completely.
If so, what about export `wapper` as `nodeWrapper`?

src/wrappers/generic/GenericExecutor.ts
serkan-ozal: Empty tag value is not a problem but the problem is empty operation name as we pass the function name as operation name below

src/index.ts
```diff
@@ -20,6 +20,7 @@ import LogManager from './plugins/LogManager';
 import InitManager from './init/InitManager';
 import ModuleUtils from './utils/ModuleUtils';
 import * as Foresight from './foresight';
+import * as GenericWrapper from './wrappers/generic/GenericWrapper';
```
serkan-ozal, line 23: Can you also add `nodeWrapper` definition into `index.d.ts`?

src/wrappers/generic/GenericExecutor.ts
```diff
@@ -0,0 +1,70 @@
...
+    const {
+        request,
+        tracer,
+    } = execContext;
+
+    const functionName = get(request, `functionName`);
+    const contextInformation: any = execContext.getContextInformation();
+    const traceId = Utils.generateId();
+
+    const rootSpan = tracer._startSpan(functionName || get(request, `__THUNDRA_ID__`), {
```
serkan-ozal, line 43:
```js
const rootSpan = tracer._startSpan(functionName || ('function-' + get(request, `__THUNDRA_ID__`))
```

serkan-ozal (approved):

PR #325, opened 2022-01-29
Change ES integration for support '@elastic/elasticsearch >= 7.0.0'

bcaglayan:
Change ES integration for support '@elastic/elasticsearch >= 7.0.0'


src/Constants.ts
```diff
@@ -566,6 +566,10 @@ export const TriggerHeaderTags = {
     RESOURCE_NAME: 'x-thundra-resource-name',
 };
 
+export const IntegrationSpecialHeaders = [
+    'x-elastic-client-meta',
```
serkan-ozal, line 570: What is this header for and when it is used?

serkan-ozal: According to here: https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/changelog-client.html#_added_x_elastic_client_meta_header_1373
> Adds the x-elastic-client-meta HTTP header which is used by Elastic Cloud and can be disabled with the enableMetaHeader parameter set to false.

So seems that it has been added since version `7.11`. So what will happen for versions until `7.11` (`7.0.x` ... `7.10.x`)?

src/integrations/ESIntegration.ts
```diff
@@ -52,38 +52,29 @@ class ESIntegration implements Integration {
      */
     wrap(lib: any, config: any) {
         ThundraLogger.debug('<ESIntegration> Wrap');
-
         function wrapRequest(request: any) {
-
             return function requestWithTrace(params: any, options: any, cb: any) {
+                if (!params || (params.method === 'GET' && params.path === '/')) {
```
serkan-ozal, line 57: Here, which requests (`GET /`) we are ignoring?

package.json
```diff
@@ -75,7 +75,7 @@
     "@babel/polyfill": "^7.0.0",
     "@babel/preset-env": "^7.0.0",
     "@babel/runtime-corejs2": "^7.0.0",
-    "@elastic/elasticsearch": "^6.2.4",
+    "@elastic/elasticsearch": "^7.0.0",
```
serkan-ozal, line 78: So, we won't test our agent against `6.x` versions right? Is it possible to test with both versions?

serkan-ozal (approved):

PR #329, opened 2022-02-23
Disable instrumenting http response body by default

bcaglayan:
Disable instrumenting http response body by default & use 'prependListener' for subscribe data event of http response object


src/config/ConfigMetadata.ts
```diff
@@ -205,7 +205,7 @@ export const ConfigMetadata: {[key: string]: { type: string, defaultValue?: any
     },
     [ConfigNames.THUNDRA_TRACE_INTEGRATIONS_HTTP_RESPONSE_BODY_MASK]: {
         type: 'boolean',
-        defaultValue: false,
+        defaultValue: true,
```
serkan-ozal, line 208: Can you also update default value for the `THUNDRA_AGENT_TRACE_INTEGRATIONS_HTTP_RESPONSE_BODY_MASK` in the `README`?

serkan-ozal (approved):

PR #370, opened 2022-10-25
Fix/planetscale node fetch auth problem

bcaglayan:
Fix Planetscale node fetch auth problem

### DIFF
--- a/src/integrations/HttpIntegration.ts
+++ b/src/integrations/HttpIntegration.ts
@@ -61,7 +61,10 @@ class HttpIntegration implements Integration {
         function wrapper(request: any) {
             return function requestWrapper(a: any, b: any, c: any) {
                 let span: ThundraSpan;
-                let { url, options, callback } = HTTPUtils.parseArgs(a, b, c);
+                const args = HTTPUtils.parseArgs(a, b, c);
+                const url = args.url;
+                let options = args.options;
+                const callback = args.callback;
                 try {
                     ThundraLogger.debug('<HTTPIntegration> Tracing HTTP request:', options);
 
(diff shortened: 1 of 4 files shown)


src/integrations/HttpIntegration.ts
```diff
@@ -58,46 +58,70 @@ class HttpIntegration implements Integration {
...
+                    const host = (
+                        (parsedUrl && parsedUrl.hostname) ||
+                        (parsedUrl && parsedUrl.host) ||
+                        (options && options.hostname) ||
+                        (options && options.host) ||
+                        (options && options.uri && options.uri.hostname) ||
+                        'localhost'
+                    );
+
+                    if (callback && callback.__thundra_wrapped) {
```
serkan-ozal, line 91: I think, you can move this logic to line `70` (just after `ThundraLogger.debug('<HTTPIntegration> Tracing HTTP request:', options);`);

src/utils/HTTPUtils.ts
```diff
@@ -157,6 +157,47 @@ class HTTPUtils {
 
         return result;
     }
+
+    static parseArgs(a: any, b: any, c: any) {
```
serkan-ozal, line 161: I think, this might be easier to read and refactor the code later by handling each case individually in if/else statements:
```js
    static parseArgs(a: any, b: any, c: any) {
        let url = a;
        let options = b;
        let callback = c;

        // Check whether first arg is URL 
        if (!(['string', 'URL'].includes(typeof a))) {
          url = a;
          if (typeof url === 'URL' || (url.constructor && url.constructor.name === 'URL')) {
              url.path = url.pathname;
          }
          if (typeof b === 'object' && !c) {
            // handling case of "got.post(url, options)"
            options = b;
            callback = undefined;
          } else if (typeof b === 'function' && !c) {
            // handling case of "request(url, callback)"
            options = undefined;
            callback = b;
          }
        } else if (typeof a === 'object' && typeof b === 'function' && !c) {
            // handling case of "request(options, callback)"
            url = undefined;
            options = a;
            callback = b;
        } 
        
        return { url, options, callback };
    }
```

serkan-ozal (approved):

PR #385, opened 2022-12-21
Try to obtain encoding from response

bcaglayan:
- try to obtain encoding from response 
- try to parse response content as json


src/utils/EncodingUtils.ts
```diff
@@ -0,0 +1,43 @@
...
+                    ThundraLogger.debug(`<EncodingUtils> Could decode data with ${encoding}`, err.message);
+                }
+            }
+
+            const decodedDataRef = decodedData;
+            try {
+                /**
+                 * Make sure data is json
+                 */
+                JSON.parse(decodedDataRef);
```
serkan-ozal, line 32: The response doesn't need to be json. It can also be xml or any other format and we are able to show non-json responses in our UI. So we should just decode if it is encoded with supported formats. Or just return `toString` of given data. 

Also another reason for removing redundant (because it is not required to have json formatted response) `JSON.parse(...)` here is that it add extra performance overhead by parsing data for most of the cases.

So the method name should be `getPayload` and `JSON.parse(...)` should be removed here.

src/utils/HTTPUtils.ts
```diff
@@ -206,6 +208,24 @@ class HTTPUtils {
...
+
+    static obtainIncomingMessageEncoding(incomingMessage: http.IncomingMessage): string {
+        if (!incomingMessage || !incomingMessage.headers || !incomingMessage.rawHeaders) {
+            return;
+        }
+
+        let contentEncoding = incomingMessage.headers
+            && (incomingMessage.headers['content-encoding']);
+        if (!contentEncoding) {
+            const encodingIndex = incomingMessage.rawHeaders.indexOf('Content-Encoding');
```
serkan-ozal, line 220: HTTP headers are case insensitive so in theory it could be sent as `content-encoding` and it is valid usage according to HTTP.

src/integrations/HttpIntegration.ts
```diff
@@ -252,7 +253,13 @@ class HttpIntegration implements Integration {
                             res.on('end', () => {
                                 try {
                                     if (chunks && chunks.length) {
-                                        const responseBody: string = Buffer.concat(chunks).toString('utf8');
+                                        const concatedChunks = Buffer.concat(chunks);
+                                        const contentEncoding = HTTPUtils.obtainIncomingMessageEncoding(res);
+                                        const responseBody: string =
```
serkan-ozal, line 258: Response body size should also be checked after decoding because decoded response size might be 4x-5x of taken encoded response size.

bcaglayan: Control added in 'getPayload' method

src/utils/HTTPUtils.ts
bcaglayan: You' re right, `incomingMessage.headers` are case insensitive (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers). Predefined header names in `incomingMessage.headers` (built-in http) are all in lower case. But in all header names in `incomingMessage.rawHeaders` look like camel case like 'Content-Encoding' (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding) .`incomingMessage.rawHeaders` is an string array which contains header names and values, For directly check Content-Encoding is exist, i used orginal header name of content encoding.

src/integrations/HttpIntegration.ts
```diff
@@ -252,11 +253,20 @@ class HttpIntegration implements Integration {
...
+                                            contentEncoding
+                                                ? (EncodingUtils.getPayload(
```
gokhan721, line 260: What if Content-Encoding have [multiple compression algorithm](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding)?

e.g: 

// Multiple, in the order in which they were applied
Content-Encoding: deflate, gzip

src/integrations/HttpIntegration.ts
```diff
@@ -252,11 +253,20 @@ class HttpIntegration implements Integration {
                             res.on('end', () => {
                                 try {
                                     if (chunks && chunks.length) {
-                                        const responseBody: string = Buffer.concat(chunks).toString('utf8');
```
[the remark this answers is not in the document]
bcaglayan: Added to HTTP/2 integration

